PR #52, opened 2021-05-01
fix: invite flow and add new batch form enhancements

pateljannat:
- [x] Add new batch form enhancements (https://github.com/fossunited/community/issues/43)
- [x] Doctype for Invite Request (https://github.com/fossunited/community/issues/42)
- [x] Portal Page for a new signup form
- [x] Invite flow from Invite Request to user creation
- [x] Widget for Request Invitation
- [x] Call to create a new invite request
- [x] Tests for Invite Request


community/www/profiles/profile.py
```diff
@@ -1,14 +1,16 @@
...
     context.no_cache = 1
     context.username = frappe.form_dict["username"]
+    print(frappe.form_dict["username"])
     context.member = get_member(context.username)
+    print(context.member)
     if not context.member:
         context.template = "www/404.html"
     else:
-        context.sketches = list(filter(lambda x: x.owner == context.member.email, get_recent_sketches()))
+        context.sketches = list(filter(lambda x: x.owner == context.member.email, Sketch.get_recent_sketches()))
```
anandology, line 13: The `get_recent_sketches` only returns the recent 100 sketches. So the number of sketches for that user could be very small or even none. The right way to do this is by adding a where clause in the db query.

Can you change the `get_recent_sketches` method to take an optional argument `owner` and add filters to `get_all` when `owner` is provided?

community/hooks.py
```diff
@@ -136,7 +136,6 @@
     {"from_route": "/sketches/<sketch>", "to_route": "sketches/sketch"},
     {"from_route": "/courses/<course>", "to_route": "courses/course"},
     {"from_route": "/courses/<course>/<topic>", "to_route": "courses/topic"},
-    {"from_route": "/hackathons", "to_route": "hackathons"},
```
anandology, line 139: Why is there a change to hackthons in this PR? Is this by mistake?

community/www/home/index.js
```diff
@@ -0,0 +1,19 @@
...
+				if (data.message == "OK") {
+					$("#invite-request-form").hide();
+					var message = `<div>
+					<p class="lead">Thanks for your interest in Mon School. We have recorded your interest and we will get back to you shortly.</p>
+					</div>`;
+					$(".jumbotron").append(message);
+				}
+			}
+		})
+	})
```
anandology, line 18: The `RequestInvite` widget is unusable without this javascript.  I think it is better to add this javascript in the widget itself. 

Also, using very generic name `#subject` is probably not a good idea.  May be `#submit-invite-request`?

community/www/courses/course.html
```diff
@@ -111,11 +111,11 @@ <h2>Your Batches</h2>
       {% endfor %}
     </div>
 
-    <a class="btn btn-primary add-batch margin-bottom" href="/add-a-new-batch?new=1&course={{course.name}}">Add a new batch</a>
+    <a class="btn btn-primary add-batch margin-bottom" href="/add-a-new-batch?new=1&course={{course.title}}">Add a new batch</a>
   {% else %}
     <div class="mentor_message">
       <p> You are a mentor for this course. </p>
-      <a class="btn btn-primary" href="/add-a-new-batch?new=1&course={{course.name}}" >Create your first batch</a>
+      <a class="btn btn-primary" href="/add-a-new-batch?new=1&course={{course.title}}" >Create your first batch</a>
```
anandology, line 118: These changes are not related to invite flow. I suggest not  to mix unrelated changes in a PR.

community/www/profiles/profile.py
```diff
@@ -8,7 +8,7 @@ def get_context(context):
     if not context.member:
         context.template = "www/404.html"
     else:
-        context.sketches = list(filter(lambda x: x.owner == context.member.email, get_recent_sketches()))
+        context.sketches = Sketch.get_recent_sketches(context.member.email)
```
anandology, line 11: Again, this is not related to invite flow.

anandology (changes requested):
Please look at the suggested changes. Also you are using tabs instead of spaces. Could you please use the editorconfig or configure your editor to replace tabs with spaces?

anandology (approved):
LGTM!

PR #53, opened 2021-05-01
feat: design homepage based on figma design

khalby786:
- The profile picture next to the course author's names doesn't actually display the profile picture. It's because I don't know what the name of the property is for getting the user's profile picture: https://github.com/khalby786/community/blob/67d3ec75c84b63b14777f38daddee2b28d68b876/community/lms/widgets/CourseTeaser.html#L10
- The header and footer is still yet to be done.

### DIFF
--- a/community/public/css/style.less
+++ b/community/public/css/style.less
@@ -1,14 +1,21 @@
 @primary-color: #08B74F;
+@import url('https://rsms.me/inter/inter.css');
+
+body {
+  font-family: "Inter", sans-serif;
+}
 
 .teaser {
   background: white;
-  border-radius: 10px;
-  border: 1px solid #ddd;
+  border-radius: 9px;
+  border: 1px solid #C4C4C4;
 }
 
 .sketch-teaser {
   .teaser();
   width: 220px;
+  margin-bottom: 30px;
+  margin-top: 30px;
 
   svg {
     width: 200px;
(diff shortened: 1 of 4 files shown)


community/public/css/style.less
```diff
@@ -1,14 +1,21 @@
 @primary-color: #08B74F;
+@import url('https://rsms.me/inter/inter.css');
```
anandology, line 2: I prefer not to have external dependencies. Could you please add the css and the fonts to `public/` folder?

community/www/home/index.html
```diff
@@ -10,7 +10,7 @@
 <section id="hero">
   <div class="container">
     <div class="jumbotron">
-      <h1 class="display-4">Guided online courses, with a mentor at your back.</h1>
+      <h1 class="display-4">Guided online courses, with a <br />mentor at your back.</h1>
```
anandology, line 13: Can you achieve this by setting a width to the display instead of <br/>.  This message may change and it would be difficult to figure out where to add the line break.

What happens when you see this on a mobile?

community/public/css/style.less
```diff
@@ -1,14 +1,21 @@
...
+body {
+  font-family: "Inter", sans-serif;
+}
 
 .teaser {
   background: white;
-  border-radius: 10px;
-  border: 1px solid #ddd;
+  border-radius: 9px;
+  border: 1px solid #C4C4C4;
```
anandology, line 11: Can you use variables instead of hard coding the colors?

anandology (changes requested):

anandology:
merged.

PR #108, opened 2021-05-24
fix: ignore user permission for membership

pateljannat:
1. LMS Batch Membership has a field member which is linked to the User doctype.
2. This link field was not considering Website Users.
3. Added a filter ignore_user_type to allow Website Users.


anandology (approved):
Looks good. Quite magical.

PR #113, opened 2021-05-28
feat: Invite based membership

### LINKED ISSUE #111: Invite Based Batch Membership
### Why?

Currently, the only way to join a batch is through the Course Details Page, where the student can view upcoming batches and join a batch that suits them. But this isn't possible if the batch is private. Private batches don't appear on the Course Details page. Mentors can add students to this batch via Desk. But not all mentors will have desk access. So such a mentor won't be able to add students to the batch.

### Solution - Invite based Batch Membership

1. We already have a select field called Membership in LMS Batch doctype with options such as Open, Restricted, Invite Only, Closed. 
2. For Batches that are private, the membership can be Invite Only. For such batches, an invite link will be visible to the mentor on the ~~Course Details~~ Batch home Page. The mentor can then share this link with the students he wants to add to this batch.
3. When a student uses the link a form will be displayed to them. The form will only ask for their email address. Batch and course information will be prefilled and read-only. On submit, membership for that batch will be created for the student. They will be redirected to the Batch home page.

### Possible Scenarios

1. What if the student who uses the link is not a user of the community? Such users can be shown a message asking them to signup before they join this batch. They can then be redirected to the signup page. For mon_school we don't have a signup page as of now. They can be redirected to the home page where they can request an invitation. Decision for mon school depends on #110

pateljannat:
1. Fixes #111 
2. Minor fixes to batch home page


community/www/batch/become-a-member/index.html
```diff
@@ -0,0 +1,73 @@
...
+    href="/login?redirect-to=/become-a-member/{{ batch.name }}">{{_("Login")}}</a>
+</div>
+
+{% elif already_a_member %}
+
+<div class="page-card">
+  <div class='page-card-head'>
+    <span class='indicator blue password-box'>Already a member</span>
+  </div>
+  <div class=''>You are already a member of the batch {{ batch.title }}.
```
anandology, line 26: Again, it is not clear which course it is. I think it is important to mention both the course title and the batch name.

community/www/batch/become-a-member/index.html
```diff
@@ -0,0 +1,73 @@
...
+</div>
+
+{% else %}
+
+<div class="page-card">
+  <div class='page-card-head'>
+    <span class='indicator blue password-box'>Confirm your membership</span>
+  </div>
+  <div>Please provide your confirmation to be a part of the batch {{ batch.title }} for the course
+    {{ batch.course_title }}.
```
anandology, line 38: 👍🏽

anandology (approved):
looks good to me. However, I found that there are some places where the message is not every clear and need to be improved. 

I think you can merge this PR and address those concerns after that.

PR #117, opened 2021-06-02
feat: lesson progress

### LINKED ISSUE #114: Show course progress to the students
Provide a way for a student to see their own progress of the course. 

We could either has a separate progress page or show the progress on the home page itself. The latter option feels like a better way as the student doesn't have to go to a seperate page to see the progress.

Since we already have a course outline, we could display it right there. I propose showing the progress in three levels.
-  show a tick for the completed lesson 
- show a progress bar for each chapter and a tick if the chapter is complete
- show a progress bar for the entite course

The progress bar for the course could be added to the course teaser pages whereever they are displayed.

pateljannat:
Closes #114


community/lms/doctype/lesson/lesson.py
```diff
@@ -43,3 +43,54 @@ def get_prev(self):
...
+
+@frappe.whitelist()
+def save_progress(lesson):
+    if frappe.db.exists("LMS Course Progress",
+            {
+                "lesson": lesson,
+                "owner": frappe.session.user
+            }):
+        return
+
```
anandology, line 63: This API can be called by any user, who is not even a member any batch. I don't think we should allow that. We need to have a check to make sure that won't happen.

community/lms/doctype/lesson/lesson.py
```diff
@@ -43,3 +43,54 @@ def get_prev(self):
...
+        "doctype": "LMS Course Progress",
+        "lesson": lesson_details.name,
+        "status": status
+    }).save(ignore_permissions=True)
+
+def update_progress(lesson):
+    user = frappe.session.user
+    if not all_dynamic_content_submitted(lesson, user):
+        return
+    course_progress = frappe.get_doc("LMS Course Progress", {"lesson": lesson, "owner": user})
```
anandology, line 85: What if this document is not found?

community/lms/doctype/lesson/lesson.py
```diff
@@ -43,3 +43,54 @@ def get_prev(self):
...
+    course_progress.status = "Complete"
+    course_progress.save()
+
+def all_dynamic_content_submitted(lesson, user):
+    exercises = frappe.get_all("Exercise", {"lesson": lesson}, ["name"])
+    all_exercises_submitted = True
+    for exercise in exercises:
+        if not frappe.db.count("Exercise Submission", {"exercise": exercise.name, "owner": user}):
+            all_exercises_submitted = False
```
anandology, line 94: This is in efficient. You should use a `IN` clause and let SQL do the hard work instead of making n queries. 

```
    query = {
        "exercise": ["IN", exercise_names],
        "owner": user
    }
    return frappe.db.count("Exercise Submission", query) == len(exercise_names)
```

community/lms/doctype/exercise/exercise.py
```diff
@@ -55,5 +56,9 @@ def submit(self, code):
             image=image,
             solution=code)
         doc.insert(ignore_permissions=True)
+
+        if not (course.is_mentor(frappe.session.user) or frappe.flags.in_test):
+            update_progress(self.lesson)
```
anandology, line 61: I guess you need to call `update_progress` on `ExerciseSubmission.before_save` and not  here.

community/lms/doctype/lesson/lesson.py
```diff
@@ -43,3 +43,54 @@ def get_prev(self):
...
+    lesson_details = frappe.get_doc("Lesson", lesson)
+    dynamic_content = frappe.db.count("LMS Section",
+                        filters={
+                            "type": ["not in", ["example", "text"]],
+                            "parent": lesson_details.name
+                        })
+
+    status = "Complete"
+    if dynamic_content:
+        status = "Partially Complete"
```
anandology, line 73: Shouldn't you call `all_dynamic_content_submitted` function to check if all the execises are complete?

pateljannat: Here I just check if there is dynamic content in the lesson, submission is checked only through update_progress

anandology (approved):

PR #158, opened 2021-07-19
fix: remove references to old styles in the hooks

anandology:
Old styles were removed in #157, but a reference to it was hanging
around in the hooks.


anandology:
My mistake. The community.bundle is still required.

PR #169, opened 2021-08-06
fix: Username validations

### LINKED ISSUE #166: Issues with Signup
Community app expects every user to have a username, but the default signup flow of frappe doesn't ask for username.

One way to fix this issue is to make sure we assign a username automatically based on the first name. We could add a number as suffix if we get into a username conflict. 

But this requires the ability to change the username after the account is created and it would be good to have a link from the profile page to do this.

There are couple of other issues with the usernames.
- We allow space character in usernames and that creates issues. I think we can follow github's policy for usernames.

```
Username may only contain alphanumeric characters or single hyphens, and cannot begin or end with a hyphen.
```

- Also, we should make sure we don't allow two usernames which vary only in the case. MySQL will compares strings without considering case, but postgres does consider case. This is something it keep a watch on.

pateljannat:
Closes #166


community/overrides/user.py
```diff
@@ -2,9 +2,50 @@
...
 class CustomUser(User):
 
+    def validate(self):
+        super(CustomUser, self).validate()
+        self.validate_username_characters()
+
+    def validate_username_characters(self):
+        if self.is_new():
+
+            if self.username.find(" "):
```
anandology, line 18: What id self.usename is `None`?

community/overrides/user.py
```diff
@@ -2,9 +2,50 @@
...
+
+            if not self.username:
+                self.username = self.get_username_from_first_name()
+
+            if self.username_exists():
+                self.username = self.remove_illegal_characters() + str(random.randint(0, 99))
+
+        else:
+            if not re.match("^[A-Za-z0-9_-]*$", self.username):
+                frappe.throw(_("Username can only contain alphabets, numbers, hyphen and underscore."))
```
anandology, line 32: Allow either hypens or underscores. I don't think we need both.

community/overrides/user.py
```diff
@@ -2,9 +2,50 @@
...
+
+    def remove_illegal_characters(self):
+        username = ''.join([c for c in self.username if c.isalnum() or c in ['-', '_']])
+        while username[0] == "-" or username[len(username) - 1] == "-":
+            if username[0] == "-":
+                username = username[1:]
+            if username[len(username) - 1]:
+                username = username[:1]
+        return username
+
```
anandology, line 48: This could be repalced with `re.sub("[^\w]+", "", username).strip("_")`
assuming you are going with underscores and without hyphens. 

If you want both, then it would become: `re.sub("[^\w-]+", "", username).strip("_-")`

PR #171, opened 2021-08-09
fix: username issues

### LINKED ISSUE #166: Issues with Signup

pateljannat:
1. Keeping only underscore and not hyphen as Frappe sets the default username with user score after signup.
2. Added validation message if the username exists if the user is not new.
3. Replaced complex code with simpler code.

Closes #166

### DIFF
--- a/community/overrides/user.py
+++ b/community/overrides/user.py
@@ -13,16 +13,23 @@ class CustomUser(User):
         self.validate_username_characters()
 
     def validate_username_characters(self):
-        if self.is_new():
+        if len(self.username):
+            underscore_condition = self.username[0] == "_" or self.username[-1] == "_"
+        else:
+            underscore_condition = ''
 
+        if self.is_new():
             if not self.username:
                 self.username = self.get_username_from_first_name()
 
             if self.username.find(" "):
                 self.username.replace(" ", "")
 
-            if not re.match("^[A-Za-z0-9_]*$", self.username):
+            if not re.match("^[A-Za-z0-9_]*$", self.username) or underscore_condition:
                 self.username = self.remove_illegal_characters()
+
+            if len(self.username) < 4:
+                self.username = self.email.replace("@", "").replace(".", "")
 
             while self.username_exists():
                 self.username = self.remove_illegal_characters() + str(random.randint(0, 99))
@@ -34,8 +41,11 @@ class CustomUser(User):
             if not re.match("^[A-Za-z0-9_]*$", self.username):
                 frappe.throw(_("Username can only contain alphabets, numbers and unedrscore."))
 
-            if self.username[0] == "_" or self.username[len(self.username) - 1] == "_":
+            if underscore_condition:
                 frappe.throw(_("First and Last character of username cannot be Underscore(_)."))
+
+            if len(self.username) < 4:
+                frappe.throw(_("Username cannot be less than 4 characters"))
 
     def get_username_from_first_name(self):
         return frappe.scrub(self.first_name) + str(random.randint(0, 99))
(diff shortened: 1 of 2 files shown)


community/overrides/user.py
```diff
@@ -15,36 +15,33 @@ def validate(self):
...
-                frappe.throw(_("Username can only contain alphabets, numbers, hyphen and underscore."))
+            if not self.username:
+                frappe.throw(_("Username already exists."))
+
+            if not re.match("^[A-Za-z0-9_]*$", self.username):
+                frappe.throw(_("Username can only contain alphabets, numbers and unedrscore."))
 
-            if self.username[0] == "-" or self.username[len(self.username) - 1] == "-":
-                frappe.throw(_("First and Last character of username cannot be Hyphen(-)."))
+            if self.username[0] == "_" or self.username[len(self.username) - 1] == "_":
```
anandology, line 37: In Python `foo[-1]` is a shorthand for `foo[len(foo)-1]`.

community/overrides/user.py
```diff
@@ -15,36 +15,33 @@ def validate(self):
...
         if self.is_new():
 
+            if not self.username:
+                self.username = self.get_username_from_first_name()
+
             if self.username.find(" "):
                 self.username.replace(" ", "")
 
             if not re.match("^[A-Za-z0-9_]*$", self.username):
                 self.username = self.remove_illegal_characters()
```
anandology, line 25: If the username is provided by the user, then you should given an error instead of silently removing the invalid characters. 

If the username is generated by the system, they it could have removed the invalid characters at that time and there no need for this check.

community/overrides/user.py
```diff
@@ -15,36 +15,33 @@ def validate(self):
...
 
-            if not self.username:
-                self.username = self.get_username_from_first_name()
-
-            if self.username_exists():
+            while self.username_exists():
                 self.username = self.remove_illegal_characters() + str(random.randint(0, 99))
```
anandology, line 28: Interesting bug here. It is possible to make the system not accept signups with any given name. If I create 101 accounts with firstname "Foo", then the the system will go into infinite loop for the next person with first name "Foo".

anandology: What if first name of a user is a single letter? Some people use abbreviations for the first name. The username generated will be an invalid username. It is good to make sure that the system will never generate an invalid username.

community/overrides/user.py
pateljannat: The system generates the username using the first_name. If during signup a user adds any illegal character to their name, it reflects in the username too. So this check is needed.

anandology: > The system generates the username using the first_name. If during signup a user adds any illegal character to their name, it reflects in the username too. So this check is needed.

I'm not debating whether or not the check is needed. I'm saying that the check is in the wrong place. 

If a function is taking the responsibiltiy of generating a username, then it should generate a valid username.

pateljannat: It would be. The only check missing was for username shorter than 4 characters. I have added a condition for the same.

PR #173, opened 2021-08-10
fix: sketch cards

pateljannat:
Added a widget for sketch card


anandology:
I don't think `SketchCard` belongs to community at all. It should be in Mon School. 

The community app doesn't even know that there exists a doctype called Sketch. Let's generalize the widgets functionality and move this widget to Mon School.

PR #272, opened 2021-11-30
Added support for Cohorts

### LINKED ISSUE #271: Add support for cohorts
**Is your feature request related to a problem? Please describe.**

Add support for cohort based courses to allow running the courses as cohorts. 

We've tried with batches earlier, but that was not very effective.

**Describe the solution you'd like** 

Implement new doctypes for cohorts and update the existing ones as required.

As per the discussion with @pateljannat, these are the doctypes that are going to be added/updated.

```
Cohort
- course
- title
- instructor
- status: Scheduled | Live | Completed | Cancelled
- begin_date
- end_date
- duration (eg. 4 weeks)

Cohort Subgroup (replace Batch?)
- cohort: Cohort
- title: str
- invite_code: str

Cohort Mentor (can we reuse/replace "LMS Course Mentor Mapping"?)
- cohort: Cohort
- email: User
- primary_subgroup: Cohort Subgroup

Cohort Staff
- cohort: Cohort
- email: User
- role: Admin | Manager | Staff

LMS Batch Membership (Rename to Course Student)
- course
- batch
- member_email: User
- member_type
- role
- cohort
- subgroup

Cohort Join Request
- cohort
- subgroup
- email
- status: Pending | Accepted | Rejected
- notes
```

**Describe alternatives you've considered**

Implementing these doctypes in mon_school, but I thought it would be good to be part of the school so that other instances can also benefit from it.

anandology:
Closes #271


school/lms/doctype/cohort/cohort.json
```diff
@@ -0,0 +1,142 @@
...
+ "index_web_pages_for_search": 1,
+ "links": [
+  {
+   "group": "Links",
+   "link_doctype": "Cohort Subgroup",
+   "link_fieldname": "cohort"
+  },
+  {
+   "group": "Links",
+   "link_doctype": "Cohort Student",
```
pateljannat, line 110: This Doctype is missing. Is this in Mon School? It's causing an error when I tried to make changes in Cohort Subgroup

pateljannat (changes requested):

school/lms/doctype/cohort/cohort.json
```diff
@@ -0,0 +1,142 @@
...
+  {
+   "fieldname": "duration",
+   "fieldtype": "Data",
+   "label": "Duration"
+  },
+  {
+   "fieldname": "slug",
+   "fieldtype": "Data",
+   "in_list_view": 1,
+   "label": "Slug",
```
pateljannat, line 63: Why are we using slug when we can use a series? To a new user, this field won't make much sense as to what data is expected here. Same for slug in cohort subgroup.

school/lms/doctype/cohort_subgroup/cohort_subgroup.json
```diff
@@ -0,0 +1,99 @@
+{
+ "actions": [],
```
pateljannat, line 2: If things with the subgroups are going well can we migrate this to LMS Batch before merging?

school/lms/doctype/cohort/cohort.json
anandology: No, this should go away. Not sure how migration was successful on mon school. I'll remove it.

school/lms/doctype/cohort/cohort.json
anandology: Slug is part of the URL. I would like to URLs to be meaningful.  For example, this is URL that we use now.

https://mon.school/courses/the-joy-of-programming/cohorts/th-2021-12

We could auto create the slug from not title, if not specified, but I would like to have an option to specify it.

Same with subgroups. Using names with case variations and spaces doesn't play well with URL and things break when we copy the URL and share it in email. Again, we could auto generate the slug from title.

school/lms/doctype/cohort_subgroup/cohort_subgroup.json
anandology: I would prefer to do that after this PR is merged.

school/lms/doctype/cohort/cohort.json
anandology: Resolved this issue in eb59713b656fc33106dd792406bdb0d42f6575d6.

school/lms/doctype/cohort/cohort.json
pateljannat: > but I would like to have an option to specify it

Why so when we can auto-create the slug from the title as you mentioned.

anandology: For the course we don't have slug because, the name itself is the slug and the name could be edited. Where as for subgroup (and also for cohort) the name can not be derived from the title along as the same college can be part of two cohorts, so we use the autogenerated name and use a slug. The combination  of course, cohort-slug and subgroup-slug forms the unique URL.

pateljannat:
@anandology any reason why this PR is in draft?

anandology:
> @anandology any reason why this PR is in draft?

I've removed the draft status now.

PR #295, opened 2022-01-25
feat: pass context to page extensions

anandology:
Pass the context to page extensions to allow them to make decisions
based on the context. For example, an extension to load the course
specific scripts. This is currently not possible because the course
details are not availale to the page extensions. Made this possible by
passing the context to page extensions.


anandology:
This is required for Mon School to add custom scripts for Self Hosting 101 course.

pateljannat (approved):

PR #572, opened 2023-07-27
fix: course when instructor is not set

tahir-zaqout:
instructors = frappe.get_all(
		"Course Instructor", {"parent": course}, ["instructor"], order_by="idx"
	)
it return like
	[
		{"instructor": "user1"},
		{"instructor": "user2"}
	]
and the
if not instructors:
    instructors = frappe.db.get_value("LMS Course", course, "owner").split(" ")

return ["user1", "user2"]

and it make error if get all return []
because instructor.instructor at  https://github.com/frappe/lms/blob/main/lms/lms/utils.py#L167

### DIFF
--- a/lms/lms/utils.py
+++ b/lms/lms/utils.py
@@ -156,7 +156,7 @@ def get_tags(course):
 def get_instructors(course):
 	instructor_details = []
 	instructors = frappe.get_all(
-		"Course Instructor", {"parent": course}, ["instructor"], order_by="idx"
+		"Course Instructor", {"parent": course}, ["instructor"], order_by="idx", pluck="instructor"
 	)
 	if not instructors:
 		instructors = frappe.db.get_value("LMS Course", course, "owner").split(" ")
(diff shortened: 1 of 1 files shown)


ahmadRagheb:
Great Job @tahir-zaqout

tahirAnvil:
@pateljannat

tahir-zaqout:
@pateljannat

lms/lms/utils.py
```diff
@@ -156,15 +156,15 @@ def get_tags(course):
 def get_instructors(course):
 	instructor_details = []
 	instructors = frappe.get_all(
-		"Course Instructor", {"parent": course}, ["instructor"], order_by="idx"
+		"Course Instructor", {"parent": course}, ["instructor"], order_by="idx", pluck="instructor"
```
niraj2477, line 159: No of pluck here. Below change would do

niraj2477:
This PR will fix : https://github.com/frappe/lms/issues/587

lms/lms/utils.py
pateljannat, line 159: @tahir-zaqout no need to have a fields array when you are plucking.

pateljannat (changes requested):

lms/lms/utils.py
pateljannat: Also please install pre-commit on your system. It will fix the linters before committing.

PR #573, opened 2023-07-28
Update certificate.py

MaiIsmail:
Because Of there is no default print format for LMS Certificate Doctype, the following error appears
image
..
So We Need to Create a Property Setter to Set Default Print Format For LMS Certificate Doctype


ahmadRagheb:
great job @MaiIsmail

NagariaHussain:
I don't think this is the right way to solve this issue, please log an issue if it still persists, closing this for now.

PR #574, opened 2023-07-29
Allow to Set Different Certificate Print Format for each LMS Course

Rola-anvil:
This is to allow the system to create different certificates for each course


ahmadRagheb:
nice job rola

pateljannat:
@Rola-anvil I don't understand the need for a link field in the LMS Course doctype.

When creating a certificate, it considers the default print format set.

You can just make your desired format as default.

Rola-anvil:
But Every course needs a different certificate design, so I have created this field to let the system user choose a custom certificate design or  consider the default one

lms/fixtures/custom_field.json
```diff
@@ -2012,5 +2012,58 @@
...
+    "bold": 0,
+    "collapsible": 0,
+    "collapsible_depends_on": null,
+    "columns": 0,
+    "default": null,
+    "depends_on": null,
+    "description": null,
+    "docstatus": 0,
+    "doctype": "Custom Field",
+    "dt": "LMS Course",
```
pateljannat, line 2028: @Rola-anvil why add a custom field in LMS Course doctype when you can add a standard field.

pateljannat (changes requested):

lms/fixtures/custom_field.json
pateljannat: Also, you have made the changes in the main branch of your fork. Can you create another branch for this feature instead?

PR #623, opened 2023-09-26
Address can't be constant as it is an entry in the form

idmlabhishek:


lms/www/billing/billing.html
```diff
@@ -66,7 +66,7 @@
 {{ super() }}
 <script src="https://checkout.razorpay.com/v1/checkout.js"></script>
 <script>
-    const address = {{ address if address else 0 }};
+    address = {{ address if address else 0 }};
```
pateljannat, line 69: The address needs to be constant. This won't fix the issue.

idmlabhishek: Hey @pateljannat  : 
A const address  throws an error when you fill the billing details. We try to fill the address in the line 
```  
const generate_payment_link = (e)=>{
                address = this.billing.get_values();
```

and at this line the assumption is that address is a var. 
Is there a better fix for this which I miss ?

PR #645, opened 2023-10-13
fix: 10th lesson access issue

### LINKED ISSUE #644: Course sidebar navigation breaks when Course Lessons are more than 9
**Describe the bug**
We are using LMS internally to host training courses. We recently came across a bug which breaks the sidebar once the course lessons are more than 9. If you look at the attached image, you will notice that the first lesson and the last lesson are highlighted. Clicking on the last lesson also simply reloads the first lesson. I also added a video of the problem below.

Screenshot 2023-10-13 at 17-24-57 Punch Presses - Metal

https://github.com/frappe/lms/assets/818803/13407313-47e8-409f-9e24-b68d792eca9d

Notice that when I hover the first and last lesson, the URL is the same. When I click either, obviously it loads the first lesson.

**To Reproduce**
1. Create a course
2. Create a chapter and link it to the course
3. Create at least 10 course lessons and link them to the chapter
4. Visit the lesson page in the browser
5. The link to the first lesson will be the same as that of the 10th lesson. The link of the 2nd lesson will be the same as that of the 11th lesson, etc

**Diagnosis**
The problem comes from the `get_lesson_details` function

```(python)
def get_lesson_details(chapter):
	lessons = []
	lesson_list = frappe.get_all(
		"Lesson Reference", {"parent": chapter.name}, ["lesson", "idx"], order_by="idx"
	)

	for row in lesson_list:
		lesson_details = frappe.db.get_value(
			"Course Lesson",
			row.lesson,
			[
				"name",
				"title",
				"include_in_preview",
				"body",
				"creation",
				"youtube",
				"quiz_id",
				"question",
				"file_type",
				"instructor_notes",
			],
			as_dict=True,
		)
		lesson_details.number = flt(f"{chapter.idx}.{row.idx}")
		lesson_details.icon = get_lesson_icon(lesson_details.body)
		lessons.append(lesson_details)
	return lessons
```

By casting the numeric strings to floats, some of the information is lost. For example, the 10th course lesson will be something like "1.10". However, when "1.10" is cast to float, you get `1.1`. That's the reason why the 1st lesson and the 10th lesson end up with the same link.

The bug exists up till commit a02365c

tundebabzy:
Fixes #644

### DIFF
--- a/lms/www/batch/learn.py
+++ b/lms/www/batch/learn.py
@@ -105,6 +105,57 @@ def get_page_extensions(context):
 		e.set_context(context)
 	return extensions
 
+class LessonBookmark:
+	"""
+	This represents a simple data structure to represent a lesson bookmark.
+	"""
+	def __init__(self, lesson_number: str) -> None:
+		self.__test_param_or_raise_exception(lesson_number)
+		_lesson_number = f"{lesson_number}." if not "." in lesson_number else lesson_number
+		first, second = _lesson_number.split(".")
+
+		self.__value = (int(first), int(second) or 0)	# second would be "" if `lesson_number` is something like "7"
+
+	def __eq__(self, other: object) -> bool:
+		if isinstance(other, self.__class__):
+			return self.__value == other.value
+		return NotImplemented
+	
+	def __gt__(self, other: object) -> bool:
+		if isinstance(other, self.__class__):
+			return self.__value > other.value
+		return False
+
+	def __repr__(self) -> str:
+		return f"{self.__value[0]}.{self.__value[1]}"
+	
+	def __test_param_or_raise_exception(self, string: str) -> None:
+		"""
+		Tests that a given string is in the format "n" or "n.n" where n is a numeric
+		character. If the string does not match the expected format, `TypeError` is
+		raised.
+		"""
+		import re
+		import sys
+		expected_format = r'^\d+\.?\d*$'
+
+		try:
+			if not re.match(expected_format, string):
+				raise TypeError("""Expected a 'str' in the format 'n' or 'n.n' where n 
+				   is a numeric character. Example: '7' or '7.10""")
+		except TypeError as e:
+			tb = sys.exc_info()[2]
+			raise TypeError("""Expected a 'str' in the format 'n' or 'n.n' where n 
+				   is a numeric character. Example: '7' or '7.10""").with_traceback(tb)
+	
+	@property
+	def value(self):
+		return self.__value
+	
+	@property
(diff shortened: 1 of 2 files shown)


tundebabzy:
This is a fix for #644

lms/www/batch/learn.py
```diff
@@ -105,12 +105,63 @@ def get_page_extensions(context):
...
+	def value(self):
+		return self.__value
+	
+	@property
+	def readable_value(self):
+		return self.__repr__()
+
 
 def get_neighbours(current, lessons):
-	current = flt(current)
```
pateljannat, line 110: @tundebabzy only removing this line is enough. What is the need of this bookmark class? According to me this file online needs this line removed and no other change.

pateljannat (changes requested):

lms/www/batch/learn.py
tundebabzy: @pateljannat, do you mean removing the cast to float using `flt`? If yes, that doesn't solve the issue with how python sorts strings. Sorting `["1.1", "2.1", "12.1"]` will result in `["1.1", "12.1", "2.1"]`.

The proper way is actually to store the "bookmark" as a tuple of `int`s like `[(1,1),(1,12), (12,1)]` which python can sort as expected

The `Bookmark` class is a data structure to abstract this and make the code more readable.

pateljannat: > that doesn't solve the issue with how python sorts strings

You are right, this won't solve the problem. I think your approach to converting it into a tuple will fix the issue. 

> make the code more readable.

I'm afraid I have to disagree with this part. Converting the numbers array into tuples and sorting them can be done in fewer lines of code.

```
def get_neighbours(current, lessons):
	numbers = [lesson.number for lesson in lessons]
	tuples_list = [tuple(map(int, s.split('.'))) for s in numbers]
	sorted_tuples = sorted(tuples_list)
	sorted_numbers = ['.'.join(map(str, t)) for t in sorted_tuples]
	index = sorted_numbers.index(current)

	return {
		"prev": sorted_numbers[index - 1] if index - 1 >= 0 else None,
		"next": sorted_numbers[index + 1] if index + 1 < len(sorted_numbers) else None,
	}
```

tundebabzy: Fewer lines in not necessarily more readable.

I was actually more concerned with the overhead and performance penalty of the class abstraction since its like 50% slower than without the abstraction even though negligible for real life scenarios. I will remove the abstraction as requested

PR #660, opened 2023-10-26
fix: timetable validation

saadchaudharry:
fix: typeError in validate_timetable

image

getting error :- **TypeError: '<' not supported between instances of 'str' and 'datetime.timedelta'
Possible source of error: mb_fxlh (app)**

Used case :-  Adding row in lms batch timetable from other doctype:
image


pateljannat:
@saadchaudharry please install pre-commit on your local setup and fix the linters.

Issue #714, opened 2024-01-03
Error sending  assignment submission  


mr8bit:
**Describe the bug**
Error sending  assignment Submission  

**To Reproduce**
1. Create lesson with assignment
2. Go to lesson and submit assignment  form
3. See error

**Expected behavior**
The answer to the task was saved and was successfully removed during verification

**Screenshots**
<img width="962" alt="Снимок экрана 2024-01-03 в 04 18 25" src="https://github.com/frappe/lms/assets/17060976/11bbc5bf-4925-4289-a1f5-b4dc4933ccc9">
<img width="1082" alt="Снимок экрана 2024-01-03 в 04 25 13" src="https://github.com/frappe/lms/assets/17060976/f15b5021-70c9-49be-ae2e-f355dff5b2db">

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Safari

mr8bit:
If you look at the code that sends for verification
<img width="806" alt="Снимок экрана 2024-01-03 в 04 26 07" src="https://github.com/frappe/lms/assets/17060976/fcac6514-24c7-4c80-8433-1c2f3932bab1">
then it does not even pass the assignment value, which is later needed to obtain the type of document to send
<img width="685" alt="Снимок экрана 2024-01-03 в 04 27 22" src="https://github.com/frappe/lms/assets/17060976/506c1cd1-7b3c-44df-b22f-d57346964474">
And this is where we get the error

mr8bit:
Can you explain how Assignment Submission was conceived?

mr8bit:
Found a similar problem on the forum
https://discuss.frappe.io/t/assignment-type-assignment-details-type-attributeerror-nonetype-object-has-no-attribute-type/111157

mr8bit:
https://github.com/frappe/lms/pull/715

pateljannat:
The entire assignment flow has been refactored recently and all these issues have been fixed with it. #417

PR #715, opened 2024-01-07
 Added worker Assignment and add More Submission type

mr8bit:
- Added more types for uploading Jupyter Notebook, Python Code, Power Point, DrawIO
- Implemented sending submission with Text and URL


mr8bit:
https://github.com/frappe/lms/issues/714

pateljannat:
@mr8bit can you please fix the linters?

mr8bit:
@pateljannat Ok, I'll fix it soon

pateljannat:
@mr8bit This version of the app will soon be replaced by a new version which is currently the develop branch of the app. The new version is significantly different. The frontend code is written in VueJs and the UI has also changed a lot. Would you be interested in contributing this feature in the new version instead of here as we don't want to add new features to the old version anymore?

PR #722, opened 2024-01-20
Update docker-compose.yml

danigbaameiro:
With the latest version of "docker-compose", you get the following error using your file "docker-compose.yml".

image

I have update the file "docker-compose.yom" in order to deploy the software with Docker (and without any error).

## Changes and Additions:

- Removed the name field at the top as docker-compose doesn't support a name field at the root level.
- Quoted the port mappings in frappe service for consistency.
- Added a depends_on section under frappe to ensure it waits for mariadb and redis to be ready.
- Introduced a custom network lms-network to facilitate communication between the containers.
- Removed the --skip-innodb-read-only-compressed command from mariadb as it might not be relevant for newer versions and can cause issues. However, if you specifically need this for your setup, you should keep it.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,5 +1,4 @@
 version: "3.7"
-name: lms
 services:
   mariadb:
     image: mariadb:10.6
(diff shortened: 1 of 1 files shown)


taowang1993:
Do you know how to deploy frappe lms on a cloud server? 

This compose file doesn't even include nginx or caddy.

PR #731, opened 2024-01-25
Update docker-installation.md

kerns:
Fixed typo.


pateljannat:
@kerns this file is no longer being used directly by the Readme and so can be deleted.

PR #756, opened 2024-03-21
Minor mod to help text.

dlaufen:
doc description update


pateljannat:
@dlaufen the macro used in the lessons has been deprecated in the new version. So this help text won't be needed anymore.

PR #776, opened 2024-04-16
fix: title and tab button copy

NagariaHussain:
Two problems with current copy: 

* "All Course" header is shown even when you switch to other tables, changes it to just Courses
* "Live" is internal terminology, "All" makes more sense

Before:
CleanShot 2024-04-16 at 14 43 13@2x

After:
CleanShot 2024-04-16 at 14 41 21@2x

### DIFF
--- a/frontend/src/pages/Courses.vue
+++ b/frontend/src/pages/Courses.vue
@@ -5,7 +5,7 @@
 		>
 			<Breadcrumbs
 				class="h-7"
-				:items="[{ label: __('All Courses'), route: { name: 'Courses' } }]"
+				:items="[{ label: __('Courses'), route: { name: 'Courses' } }]"
 			/>
 			<div class="flex">
 				<router-link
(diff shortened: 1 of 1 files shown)


NagariaHussain:
On second thought, all is really not all for a course creator, maybe it should be. I see need for a "badge" on card.. like "Upcoming", "New", etc. wdyt @pateljannat ?

pateljannat:
@NagariaHussain **All** would be a wrong label for the first tab. It shows only the published (Live) courses.

Hence the title of the page is **All Courses**, as it then shows all the tabs with various categories of courses.

I think the labels should be kept as is for now.

I don't think there is a need for a badge on the card, the cards already have too much information. You can use the tags to show New or Upcoming.

NagariaHussain:
How about changing upcoming to new and upcoming? 

We need some way to showcase new courses. Currently they show up at the end of the list, which is not good for discoverability IMO

pateljannat:
@NagariaHussain Yes I can add a tab for new courses where every course that was published within 3 months can be shown. Will also add a field for publishing date.

PR #785, opened 2024-04-26
Update README.md

Bowrna:
I tried to clone the branch and after that doing cd into the directory from the current location works.


pateljannat:
Ideally, the setup is run directly from the frappe-bench folder, and so after the clone, the valid path further is apps/lms. This PR is invalid.

Bowrna:
@pateljannat It was given that way under the docker way of running the LMS. when running via docker, it didn't create app folder. that's why I have raised this PR.
Screenshot 2024-05-02 at 9 29 09 PM

taowang1993:
So I have to install bench first. But the bench repo is aiming for installing erpnext instead. I am confused.

PR #790, opened 2024-04-29
fix: quiz attempts load issue

NagariaHussain:
Problem:

`attempts` resource was only getting reloaded onSuccess of quiz. But since `quiz` is cached, onSuccess does not run again on remount of the quiz (try switching between lessons). This results in max attempts exceeded message.

CleanShot 2024-04-29 at 19 36 30@2x

Solution: 
Reload attempts whenever quiz data changes


NagariaHussain:
Haven't touched server side code, but tests are failing ... unrelated.

PR #792, opened 2024-05-01
fix: check if role exists before renaming

dj12djdjs:
for whatever reason this role already existing on one of the sites I tried migrating ...


pateljannat:
@dj12djdjs fix linter by installing pre-commit

PR #794, opened 2024-05-02
feat: CertifiedParticipants.vue page is now searchable

arunmathaisk:
Got an slight ick while browsing through the cerificates page 😅, so made the page searchable.

https://github.com/frappe/lms/assets/61804369/cbfb3baa-323c-4a84-a8b1-fc874be59b4b


pateljannat:
@arunmathaisk fix linters

pateljannat (changes requested):

PR #799, opened 2024-05-05
Update docker-installation.md

Leopere:
Modernized and Clarified docker-installation.md
Removed superfluous steps and whitespace.
Updated the docker compose to the modern docker compose plugin since `docker-compose` itself is likely to get deprecated and the `docker compose plugin` is baked into docker. Implemented headers in Markdown as that is more compatible with screen readers.

### DIFF
--- a/docker-installation.md
+++ b/docker-installation.md
@@ -1,37 +1,56 @@
-**Step 1:** Clone the repo
+# Docker Installation Guide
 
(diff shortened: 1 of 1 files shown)


Leopere:
I'm just doing a couple of hit and runs but I'm open to suggestions on how better to submit this.

Issue #814, opened 2024-05-14
Student User should not allow to show New Batch & New Job


thida-bcn:
![image](https://github.com/frappe/lms/assets/62122640/52f7b1e2-f38e-4635-8c15-38001746cf38)
![image](https://github.com/frappe/lms/assets/62122640/603d0c8e-b26b-4aa0-8653-85641d198c49)
student are not desk user 
they are not authorized for creation of batch and job

PR #821, opened 2024-05-16
fix: batch related issues

### LINKED ISSUE #817: Batch to Course Link
when we click course from batch page, system should go to course
image

pateljannat:
Fixed #817 and #814


thida-bcn:
thank you
it's not working as expected

PR #849, opened 2024-05-29
feat: configure sidebar items

pateljannat:


thida-bcn:
hello, 
certification still enable to students, even we had uncheck at setting
image

https://github.com/frappe/lms/assets/62122640/8283e4a2-2a9d-44af-8812-a053011ca35e


image

PR #877, opened 2024-06-14
fix: show "Seats Left" instead of "Seat Left" if more than 1 seat is available

nikkothari22:
Kinda pointless PR but meh 🤷🏻 - it was irking me.

PS: Not sure if this is the right way to do it since the code (especially the styles) are duplicated in the two badges.


frontend/src/components/BatchCard.vue
```diff
@@ -4,7 +4,14 @@
 		style="min-height: 150px"
 	>
 		<Badge
-			v-if="batch.seat_count && batch.seats_left > 0"
+			v-if="batch.seat_count && batch.seats_left > 1"
+			theme="green"
+			class="self-start mb-2"
+		>
+			{{ batch.seats_left }} {{ __('Seats Left') }}
```
pateljannat, line 11: Maybe try with 2 spans here instead of two badges.

pateljannat (changes requested):

NagariaHussain:
CleanShot 2024-06-25 at 21 23 29@2x

😅

frontend/src/components/BatchOverlay.vue
```diff
@@ -5,7 +5,7 @@
 			theme="green"
 			class="self-start mb-2 float-right"
 		>
-			{{ seats_left }} {{ __('Seat Left') }}
+			{{ seats_left }} <span v-if="batch.seats_left > 1">{{ __('Seats Left') }}</span><span v-else-if="batch.seats_left == 1">{{ __('Seat Left') }}</span>
```
NagariaHussain, line 8: Should have been just `seats_left` and not `batch.seats_left`

PR #902, opened 2024-06-28
feat: limit questions in quiz

pateljannat:
You can now limit the number of questions to be shown in a quiz. So if a quiz has 10 questions and you want the students to attempt only 5 questions, you can set it as the limit. Students will then only see the first 5 questions when attempting the quiz.

If the Shuffle Questions setting is also enabled then students will see random 5 questions from the total questions in the quiz. Also based on the limit, the total marks get calculated for the quiz.

Screenshot 2024-06-28 at 12 32 04 PM


thida-bcn:
thanks

PR #931, opened 2024-07-15
feat: Dynamic Filters in Lms Portal

Gopal8848:
Dynamic filters for courses and batches portal pages based on link field selection in **LMS Filters** Doctype.

Screenshot from 2024-07-15 10-56-27
Screenshot from 2024-07-15 10-55-57


pateljannat:
@Gopal8848 I don't think a dynamic filter is a good idea here. This causes more chances of breaking things than improvement. The UI here does not match the standard. Plus too many options would just end up confusing users.

PR #988, opened 2024-08-23
A Quiz Lesson does not update progress after 30 seconds

hyperdatavn:
Issue1: 
- In Case: A lesson contain Quiz, if student is not completed a quiz within 30 seconds then save_progress API will not be call anymore. So when student completed a quiz, a progress still dont be updated.
- Solution:  Retry save_progress each 30 seconds until student completed a quiz.

Issue2: 
- In case: A lesson contain Quiz, if student submit and completed a quiz, it need to await 30 seconds for a interval be called and update progress. If student refresh or quit a page, a progress dont be updated. 
- Solution:  Add more note "After Submit a Quiz, please await 30 seconds to update a progress percentage" for student need to wait 30 seconds after they completed a quiz.

### DIFF
--- a/frontend/src/components/Quiz.vue
+++ b/frontend/src/components/Quiz.vue
@@ -29,6 +29,11 @@
 					).format(quiz.data.time)
 				}}
 			</div>
+			<div class="leading-relaxed">
+				{{
+					__('After Submit a Quiz, please await 30 seconds to update a progress percentage!')
+				}}
+			</div>
 		</div>
 		<div v-if="activeQuestion == 0">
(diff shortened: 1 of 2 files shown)


hyperdatavn:
> @hyperdatavn A better approach would be to call the save_progress function when a student submits a quiz. Can you please do that approach instead

I think about this solution already, but if i call save_progress when student Submit a quiz then it just update on backend, a progress of lesson still dont refresh in Lesson.vue page, how can i refresh progress in Lesson session  ?

pateljannat:
Fixed with #1223

PR #1017, opened 2024-09-16
refactor: adding quiz and uploading content to a lesson

pateljannat:
1. Previously there was a components sidebar, where creators had to first select the editor in which they wanted to add the quiz or upload a file too. Then after selecting the editor, they could add a file or quiz to the lesson.
2. As per feedback received, the UX was confusing. This PR will simplify the process. 
3. Adding any of this content will now happen through the editor menu itself. 
4. Also, to make the process even simpler, explanations along with help videos have been added to the right sidebar of the lesson editor. Here creators can check how to add a quiz, upload content, or add a YouTube video.

https://github.com/user-attachments/assets/39c267e0-3617-4433-be44-c2a3c2c734a0

https://github.com/user-attachments/assets/526df76b-622f-4c04-b601-adf17da0084b

https://github.com/user-attachments/assets/69ab72d0-2380-4138-a5a7-16d5f418c740


federicocalvo:
excellent function. By the way, what software do you use to record the videos with zoom, they are very good!

Issue #1064, opened 2024-10-14
Live Zoom class issue


ajiragroup:
1. Student can start a class ? Shouldn't there be only Join ?
![image](https://github.com/user-attachments/assets/2526cb61-6e8a-430f-94fd-712f47629d7c)

2. Why is it that we can start/join a class on future date ?

![image](https://github.com/user-attachments/assets/555a2019-87b7-488c-bb82-d9914354edcc)

Update:

3. Why are we still able to start/join a class in the past ?
![image](https://github.com/user-attachments/assets/26babf8a-292d-4ce9-bc71-f0c313f475e0)

4. Not so important, but why are we able to create classes beyond the batch'es start and end dates ?

anhlevfx:
Hi @ajiragroup aj Did you start successfully Zoom when press start or join button? I'm facing the issue click these buttons not trigger anything. I installed Zoom on my local machine.

ajiragroup:
> Hi @ajiragroup aj Did you start successfully Zoom when press start or join button? I'm facing the issue click these buttons not trigger anything. I installed Zoom on my local machine.

Yes zoom opens.

anhlevfx:
Did you tried on local machine, or must deploy to server with a ssl domain to make Zoom works?
I agree with you about buttons always show up, must be disabled before and after start time.

ajiragroup:
I tried on both local machine and on ssl enabled server. Zoom Works in both.

Hopesonjames:
hello  @ajiragroup  
Could you please guide me the steps on how to properly integrate our LMS with Zoom  on local setup

ajiragroup:
> hello @ajiragroup Could you please guide me the steps on how to properly integrate our LMS with Zoom on local setup

there is a good documentation on this. 

https://docs.frappelms.com/batch-creation/create-a-live-class.html

pateljannat:
@ajiragroup which version of the app are you on? I don't see the start button when I login as a student.

ajiragroup:
@pateljannat Looks like you fixed the bug first then checked my issue. https://github.com/frappe/lms/pull/1077

I am on 2.9.0

PR #1077, opened 2024-10-23
fix: show live class start button only to moderators and evaluators

pateljannat:
1. Live class start button should only be visible to Moderators and Evaluators.


ajiragroup:
Solves https://github.com/frappe/lms/issues/1064

PR #1086, opened 2024-10-29
fix: small bug in course_progress_summary.py

0xflotus:
This is a small logical fix.

Otherwise if `row.progress == 10 or row.progress == 40 or row.progress == 70` wouldn't have an effect.

### DIFF
--- a/lms/lms/report/course_progress_summary/course_progress_summary.py
+++ b/lms/lms/report/course_progress_summary/course_progress_summary.py
@@ -93,11 +93,11 @@ def get_charts(data):
 	for row in data:
 		if row.progress == 100:
 			completed += 1
-		elif row.progress < 100 and row.progress > 70:
+		elif row.progress < 100 and row.progress >= 70:
 			less_than_hundred += 1
-		elif row.progress < 70 and row.progress > 40:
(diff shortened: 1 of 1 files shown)


lms/lms/report/course_progress_summary/course_progress_summary.py
```diff
@@ -93,18 +93,18 @@ def get_charts(data):
...
-		elif row.progress < 40 and row.progress > 10:
+		elif row.progress < 40 and row.progress >= 10:
 			less_than_forty += 1
 		elif row.progress < 10:
 			less_than_ten += 1
 
 	charts = {
 		"data": {
-			"labels": ["0-10", "10-40", "40-70", "70-99", "100"],
+			"labels": ["0-9", "10-39", "40-69", "70-99", "100"],
```
pateljannat, line 107: @0xflotus can we change the conditions so that the brackets are 0-10, 11-40, 41-70, 71-100

pateljannat (changes requested):

lms/lms/report/course_progress_summary/course_progress_summary.py
0xflotus: sure.

PR #1107, opened 2024-11-08
fix: time field issue in edit batch form

### LINKED ISSUE #1103: time field issue in edit batch form
**Describe the bug**
Start and end time values are not showing in the batch edit form.

**To Reproduce**
1. Go to Edit batch and check Start and End time field

**Expected behavior**

**Screenshots**

https://github.com/user-attachments/assets/6436474e-94dc-4666-be80-7931edb58e8d

iamejaaz:
Closes #1103 

## Before
image

## After
image

### DIFF
--- a/frontend/src/pages/BatchForm.vue
+++ b/frontend/src/pages/BatchForm.vue
@@ -339,6 +339,13 @@ const batchDetail = createResource({
 			let key = checkboxes[idx]
 			batch[key] = batch[key] ? true : false
 		}
+		let timeFields = ['start_time', 'end_time']
+		for (let timeField of timeFields) {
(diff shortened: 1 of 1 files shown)


pateljannat:
@iamejaaz I don't understand the issue you are trying to solve here. The time field works properly currently. What exactly is the issue you are facing?

Screenshot 2024-11-12 at 11 58 28 AM

iamejaaz:
> @iamejaaz I don't understand the issue you are trying to solve here. The time field works properly currently. What exactly is the issue you are facing?

@pateljannat Please try selecting an AM time and check, as I showed the issue in the video. When I select an AM time, like 12 AM or 1 AM, it does not display in the edit form.

PR #1111, opened 2024-11-10
feat: allow same date live class creation

### LINKED ISSUE #1063: Allow same date Live class creation
While creating a live class, if user inputs same date but future time, it gives error : Please select future date.

iamejaaz:
Closes #1063 

## Reason of PR
- Allow users to create live classes on the same date with a future time, ensuring that the class time is validated relative to the selected timezone.

## Changes in PR
- Imported `timezone` and `utc` plugins from `dayjs` to handle timezone conversions and comparisons.
- Updated course time validation to ensure the class time is in the future based on the selected timezone, rather than comparing it with the system's local time.

### DIFF
--- a/frontend/src/components/Modals/LiveClassModal.vue
+++ b/frontend/src/components/Modals/LiveClassModal.vue
@@ -156,19 +156,19 @@ const submitLiveClass = (close) => {
 	return createLiveClass.submit(liveClass, {
 		validate() {
 			if (!liveClass.title) {
-				return 'Please enter a title.'
+				return __('Please enter a title.')
 			}
 			if (!liveClass.date) {
-				return 'Please select a date.'
+				return __('Please select a date.')
 			}
 			if (!liveClass.time) {
-				return 'Please select a time.'
+				return __('Please select a time.')
 			}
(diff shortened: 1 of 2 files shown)


frontend/src/components/Modals/LiveClassModal.vue
```diff
@@ -161,21 +161,30 @@ const submitLiveClass = (close) => {
...
 				return 'Please select a date.'
 			}
-			if (dayjs(liveClass.date).isSameOrBefore(dayjs(), 'day')) {
-				return 'Please select a future date.'
-			}
 			if (!liveClass.time) {
 				return 'Please select a time.'
 			}
+			if (!liveClass.timezone) {
+				return 'Please select a timezone.'
```
pateljannat, line 168: Translate all the strings here

pateljannat (changes requested):

frontend/src/components/Modals/LiveClassModal.vue
iamejaaz: The changes have been made.

PR #1168, opened 2024-12-05
fix: show role for members

### LINKED ISSUE #1126: Display user roles in members
image

I hope no explanation is needed here.

KerollesFathy:
Closes #1126 

- Before Fix
before_fix
---
- After Fix
after_fix

### DIFF
--- a/frontend/src/components/Members.vue
+++ b/frontend/src/components/Members.vue
@@ -66,7 +66,10 @@
 									<div class="text-gray-900">
 										{{ member.full_name }}
 									</div>
-									<div class="px-1" v-if="getRole(member.role)">
+									<div
+										class="px-1"
+										v-if="member.role && getRole(member.role) !== 'Student'"
+									>
 										<Badge
 											:variant="'subtle'"
 											:ref_for="true"
--- a/frontend/src/components/Members.vue
+++ b/frontend/src/components/Members.vue
@@ -66,8 +66,16 @@
 									<div class="text-gray-900">
 										{{ member.full_name }}
 									</div>
-									<div v-if="getRole(member)">
-										{{ getRole(member) }}
+									<div class="px-1" v-if="getRole(member.role)">
+										<Badge
+											:variant="'subtle'"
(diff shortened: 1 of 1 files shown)


frontend/src/components/Members.vue
```diff
@@ -66,8 +66,16 @@
 									<div class="text-gray-900">
 										{{ member.full_name }}
 									</div>
-									<div v-if="getRole(member)">
-										{{ getRole(member) }}
+									<div class="px-1" v-if="getRole(member.role)">
```
pateljannat, line 69: Let's show the role only is the User is a Moderator, Course Creator, or Batch Evaluator. The rest are assumed as Students. Too many badges will make the view cluttered.

pateljannat (changes requested):

frontend/src/components/Members.vue
KerollesFathy: The logic in the `get_members` API endpoint assigns only one role to each member based on a defined hierarchy. As a result, only a single role is displayed on the UI for each member


https://github.com/frappe/lms/blob/develop/lms%2Flms%2Fapi.py#L617

pateljannat: That is not what I meant. In a list, most users will be students, and very few will be Instructors or Moderators. Showing a badge for students would mean that most of the rows would have that same badge, which would look cluttered. Only having a badge once in a while when the member has a higher role would look cleaner.

pateljannat (changes requested):

frontend/src/components/Members.vue
pateljannat: That is not what I meant. Most of the users are going to be students. So every row in the list showing a student badge would not be so good. Instead, if only users who have higher roles have the badge, that would be more meaningful. The rest of them will be understood as students.

KerollesFathy: @pateljannat, yes, you are right. I fixed it :+1:

PR #1236, opened 2025-01-07
fix: fixed typo in spelling in frappe.utils.date_diff import

FahidLatheef:
- Latest Develop branch causes Internal Server Error due to incorrect python import.
- It should be `date_diff` instead of `days_diff` in frappe.utils.data

### DIFF
--- a/lms/lms/api.py
+++ b/lms/lms/api.py
@@ -22,7 +22,7 @@ from frappe.utils import (
 	now,
 	add_days,
 	format_date,
-	days_diff,
+	date_diff,
 )
 from typing import Optional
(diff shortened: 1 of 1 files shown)


pateljannat:
@FahidLatheef It is not a typo. https://github.com/frappe/frappe/blob/develop/frappe/utils/data.py#L333

FahidLatheef:
I did more research. This function `days_diff` doesn't exist in version-15 releases.

https://github.com/frappe/frappe/blob/version-15/frappe/utils/data.py

@pateljannat  
Reopening this PR to handle version-15 users.

PR #1282, opened 2025-02-03
feat: Enable Media Attachments in Course Lessons by Converting Attachments to JSON Format for Display on Website

Prabhjot-frappe:
In the current setup, users can add media, text, and video links directly from the website when creating courses. However, when using the Desk view of the Course doctype, there is no built-in functionality to add media files (such as video or audio) directly. While the "Lesson" doctype allows users to attach files, there was no seamless way to convert these attachments into a format that could be used directly on the website.

What We Did:

To bridge this gap, we created a solution that allows attachments added to "Course Lesson" records to be automatically converted into a JSON format. This JSON structure can then be used to display media files (such as images, videos, and documents) on the website.

Modification in Python Files: We updated the logic in the Python files to capture file uploads and convert them into a structured JSON block. This block contains details such as the file's URL and type (image, video, document, etc.).

Use of Hooks: To ensure the changes happen seamlessly, we used Frappe's hooks to trigger these updates. When a file is uploaded as an attachment in the "Course Lesson" doctype, the get_files_attached_to_lesson function is called to convert and store the attachment in the correct JSON format. Similarly, when a file is deleted, the remove_json_block_on_attachment_delete function ensures the corresponding JSON block is removed.

This approach provides a clean and efficient way for users to manage media content in lessons, ensuring that attachments are transformed into a format that can be directly used on the website for course displays.

In summary:

We enhanced the "Lesson" doctype by automatically converting file attachments into a JSON format.
This allows the uploaded media to be displayed on the website, making course content creation more versatile and user-friendly.
Hooks were leveraged to ensure these changes happen automatically without requiring additional user interaction.

[Screencast from 03-02-25 02:18:37 PM IST.webm](https://github.com/user-attachments/assets/5acf34ea-2381-4904-bb22-2b325da1ec40)

### DIFF
--- a/lms/hooks.py
+++ b/lms/hooks.py
@@ -104,6 +104,11 @@ doc_events = {
 	},
 	"Discussion Reply": {"after_insert": "lms.lms.utils.handle_notifications"},
 	"Notification Log": {"on_change": "lms.lms.utils.publish_notifications"},
+	"File": {
+	"after_insert": "lms.lms.utils.get_files_attached_to_lesson",
+	"on_trash": "lms.lms.utils.remove_json_block_on_attachment_delete"
+}
+
 }
 
 # Scheduled Tasks
--- a/lms/lms/utils.py
+++ b/lms/lms/utils.py
@@ -28,6 +28,7 @@ from frappe.utils import (
 )
 from frappe.utils.dateutils import get_period
 from lms.lms.md import find_macros, markdown_to_html
+import time
 
 RE_SLUG_NOTALLOWED = re.compile("[^a-z0-9]+")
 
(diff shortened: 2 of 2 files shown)


pateljannat:
@Prabhjot-frappe I don't think this solution is ideal. Firstly, I don't recommend adding lesson content from the desk. The portal should be used for it. Secondly, with this solution, any attachment that I add to Course Lessons becomes a content block, but what if that is just for my reference and not meant to be a part of the lesson?

sapna-kcs:
Agree @pateljannat but this is client requirement and we are unable to override the standard behavior to make this customization, hence the contribution. Is there a better way to achieve this?

pateljannat:
@sapna-kcs maybe do this via a server script on the client's site?

sapna-kcs:
Yes thats possible but on migrate, the script gets removed. Is there a better way to do it?
We could not find any medium and hence suggested this add-on contribution.

Is there a reason we do not want the backend to play a role for addl. features already available via the frontend?

pateljannat:
@sapna-kcs scripts don't get removed on migrate. If you are concerned about moving them from one site to another, then maybe create fixtures for them?

Prabhjot-frappe:
@rmehta As suggested, request your support.

rmehta:
@Prabhjot-frappe do check Jannat's concerns. How does this JSON list in course content get consumed? Why can't we create courses from the front-end? What is the real problem you are trying to solve here?

Prabhjot-frappe:
@rmehta sir This feature is a client requirement, where the administrator needs the ability to add video files and attachments from the desktop. To implement this, we used monkey patching. However, we understand that Frappe does not recommend monkey patching.

If this pull request cannot be merged, we would appreciate any suggestions or recommended alternatives to achieve this functionality in a way that aligns with Frappe’s best practices.

rmehta:
Why does the client want to add files from desk and not the course?

PR #1341, opened 2025-02-26
fix: typo

mahsem:
Backport v15


pateljannat:
@mahsem don't change the JSON, make the change in the doctype and let it reflect in the JSON. With the current change, it won't reflect in the doctype on migration.

PR #1369, opened 2025-03-10
fix: fixed issue with Lesson Render for SCORM Chapters

### LINKED ISSUE #1349: Continue Learning Button Doesn't work for SCORM chapters
If we click Continue Learning, page refreshes itself for SCORM Chapters because of invalid link. Only way to go to course is clicking the SCORM chapter from the Course Outline Session.

https://github.com/user-attachments/assets/080622c9-9c27-4e48-8d23-7b5417348a20

FahidLatheef:
Closes #1349

https://github.com/user-attachments/assets/7a267199-f675-4473-8031-e3a985c0dab6

### DIFF
--- a/lms/lms/utils.py
+++ b/lms/lms/utils.py
@@ -1260,16 +1260,27 @@ def get_lesson(course, chapter, lesson):
 	chapter_name = frappe.db.get_value(
 		"Chapter Reference", {"parent": course, "idx": chapter}, "chapter"
 	)
+	is_scorm_chapter = frappe.db.get_value(
+		"Course Chapter", chapter_name, "is_scorm_package"
+	)
+
 	lesson_name = frappe.db.get_value(
 		"Lesson Reference", {"parent": chapter_name, "idx": lesson}, "lesson"
 	)
 	lesson_details = frappe.db.get_value(
 		"Course Lesson",
 		lesson_name,
-		["include_in_preview", "title", "is_scorm_package"],
+		["include_in_preview", "title"],
 		as_dict=1,
 	)
-	if not lesson_details or lesson_details.is_scorm_package:
+
+	if is_scorm_chapter:
+		return {
+			"is_scorm_package": True,
+			"chapter_name": chapter_name,
+		}
+
+	if not lesson_details:
 		return {}
 
 	membership = get_membership(course)
--- a/frontend/src/pages/Lesson.vue
+++ b/frontend/src/pages/Lesson.vue
@@ -255,6 +255,15 @@ const lesson = createResource({
 			})
 			return
 		}
+		if (data.is_scorm_package) {
+			router.push({
+				name: 'SCORMChapter',
+				params: {
+					courseName: props.courseName,
+					chapterName: data.chapter_name,
+				},
+			})
+		}
 		lessonProgress.value = data.membership?.progress
 		if (data.content) editor.value = renderEditor('editor', data.content)
 		if (


lms/lms/utils.py
```diff
@@ -1260,16 +1260,27 @@ def get_lesson(course, chapter, lesson):
...
+		"Course Chapter", chapter_name, "is_scorm_package"
+	)
+
 	lesson_name = frappe.db.get_value(
 		"Lesson Reference", {"parent": chapter_name, "idx": lesson}, "lesson"
 	)
 	lesson_details = frappe.db.get_value(
 		"Course Lesson",
 		lesson_name,
-		["include_in_preview", "title", "is_scorm_package"],
```
pateljannat, line 1229: @FahidLatheef, why not use the is_scorm_package field from the lesson directly instead of fetching from the chapter?

pateljannat:
@FahidLatheef, can you please resolve the conflict

FahidLatheef:
> @FahidLatheef, can you please resolve the conflict

I'll get back to you once I test it again with the latest code

lms/lms/utils.py
FahidLatheef: This was the source of the bug. For Lessons in SCORM Chapters, lesson_details was coming as Falsy value. I will verify this with the latest code and get back to you.

FahidLatheef:
@pateljannat 

The issue still persist in the current develop branch.

**Current Develop.**

https://github.com/user-attachments/assets/12777792-04cc-492b-8f62-d3b7f99b31f8

**After the Fix.**

https://github.com/user-attachments/assets/266804d7-10ef-4963-960d-55c61c5cd816

lms/lms/utils.py
```diff
@@ -1252,16 +1252,27 @@ def get_lesson(course, chapter, lesson):
...
+		["include_in_preview", "title"],
 		as_dict=1,
 	)
-	if not lesson_details or lesson_details.is_scorm_package:
+
+	if is_scorm_chapter:
```
pateljannat: Instead of fetching is_scorm_package from Course Chapter, why can't we check it directly from lesson details?
```
if not lesson_details:
	return {}
if lesson_details.is_scorm_package
	return {
		"is_scorm_package": True
	}
```

FahidLatheef: > Instead of fetching is_scorm_package from Course Chapter, why can't we check it directly from lesson details?
> 


https://github.com/user-attachments/assets/5766ee0f-d1c9-4420-a5b9-99a965b6e89a

`lesson_name` is coming as `None` due to which subsequent `lesson_details` is also evaluated as `None` causing the Bug.

FahidLatheef: Upon Further Investigation, `idx` is wrongly created in Lesson Reference Child Doctype.

wrong_idx_passed_for_SCORM

pateljannat: @FahidLatheef, this seems like some data issue at your end, because idx cannot be 0. Can you make a fresh course on your local site and add a SCORM chapter to it, and verify the lesson idx there?

FahidLatheef: This function `lms.lms.api.add_lesson` (which is only used 1 time in the app) is the culprit. There is no idx set inside this function. Hence idx is being defaulted to 0.

add_lesson_no_idx_set

pateljannat: @FahidLatheef you can fix that in the same PR

FahidLatheef: What about existing DB Entries? Should I add a patch to update it?

Anjalii-Patel (approved):

lms/lms/utils.py
pateljannat: @FahidLatheef yes

PR #1377, opened 2025-03-12
fix: misc

NagariaHussain:
* Track current tab in batches and courses
* Fix trailing `?` without any query params


NagariaHussain:
@pateljannat lot of repitetion in `courses.vue` and `batches.vue`. Probably needs a round of refactoring.

PR #1404, opened 2025-03-28
feat:HLS Encrypted Video Support for Secure Content Streaming

domjv:


pateljannat:
@domjv adding such content should work directly on paste, without showing any prompt or placeholder to the users. Also, now that we have added support for CloudFlare video embed, have you tried that out?

PR #1467, opened 2025-04-25
feat: Embedding for CloudflareStream

harshpwctech:
CloudFlare Stream is another video streaming solution.

https://developers.cloudflare.com/stream/viewing-videos/using-the-stream-player/


harshpwctech:
@pateljannat - Hope this shouldn't have any issue.

PR #1472, opened 2025-04-28
Trainer visibility section added

Sneha-ss-17:


pateljannat:
@Sneha-ss-17 

1. Add description and screenshots to this PR.
2. The system doesn't have an LMS Trainer role, so what is this PR introducing?

PR #1501, opened 2025-05-13
course filteration based on students

Gokul1407:


pateljannat:
@Gokul1407, these changes don't make sense. Guests also need to see the list of courses. Returning an empty array would break the experience for them. Also, in most cases, students would want to enroll in more course and so showing a list of enrolled courses is also not okay. I am closing this PR.

Gokul1407:
I have mistakenly created a pull request to the frappe/lms instead of  my forked repository

PR #1572, opened 2025-06-11
feat: Embedding for Bunny Stream

harshpwctech:
Bunny Stream is another video streaming solution for edtechs looking to scale.

https://docs.bunny.net/docs/stream-embedding-videos


harshpwctech:
@pateljannat - Can you please help in merging this feature as well to the core...

PR #1601, opened 2025-06-27
chore: upgrade frappe-ui to the latest version and fixed #1587

### LINKED ISSUE #1587: toggle switch not moving
Image

Its on but the switch is set to off. only the icon is bright which tells me it is on. also check the off icon

addeeandra:
Before: the switch indicator doesn't translated properly when turned on (check #1587)

After a quick check, I saw that the code directly use `Switch` from `frappe-ui` component. Thus it must be fixed in the `frappe-ui`, but after upgrading to latest version, it seems the issue already solved.

### After `frappe-ui` upgraded

| Settings > General | Settings > Sidebar |
| --- | --- |
| image | image |

Let me know if there's any concerns about upgrading `frappe-ui`.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -31,7 +31,7 @@
     "codemirror": "^6.0.1",
     "dayjs": "^1.11.6",
     "feather-icons": "^4.28.0",
-    "frappe-ui": "^0.1.147",
+    "frappe-ui": "^0.1.80",
     "highlight.js": "^11.11.1",
     "lucide-vue-next": "^0.383.0",
     "markdown-it": "^14.0.0",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -31,7 +31,7 @@
     "codemirror": "^6.0.1",
     "dayjs": "^1.11.6",
     "feather-icons": "^4.28.0",
-    "frappe-ui": "^0.1.80",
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -31,7 +31,7 @@
     "codemirror": "^6.0.1",
     "dayjs": "^1.11.6",
     "feather-icons": "^4.28.0",
-    "frappe-ui": "^0.1.147",
+    "frappe-ui": "^0.1.80",
```
pateljannat: Seems like you have downgraded it 👀

addeeandra: Haha, my baddd, sorry 😂

I clicked from google search and it gave me 0.1.80 as its default active version, which I thought the latest version. 👀
 https://www.npmjs.com/package/frappe-ui/v/0.1.80?activeTab=versions

Gonna check into this again. Thanks! 😅

addeeandra:
Regarding this issue, it seems okay with the previous `frappe-ui` version, but it only requires upgrade on `yarn.lock`. I suggest to check and test it out thoroughly, it may introduce breaking changes which I'm not familiar with.

pateljannat:
Closed with https://github.com/frappe/lms/pull/1606

PR #1607, opened 2025-06-30
fix: add_context_for_State

mahsem:
Backport main

### DIFF
--- a/frontend/src/pages/Billing.vue
+++ b/frontend/src/pages/Billing.vue
@@ -80,7 +80,7 @@
 							/>
 							<FormControl :label="__('City')" v-model="billingDetails.city" />
 							<FormControl
-								:label="__('State')"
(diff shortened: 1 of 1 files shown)


pateljannat:
@mahsem, what issue is this PR trying to solve? Please add a proper description and reopen again.

mahsem:
@pateljannat State in Frappe is Workflow State and here it is Address .To translate it to other languages we need context otherwise LMS translation will overwrite translation for Frappe
Here are patches for frappe https://github.com/frappe/frappe/pull/32533
and erpnext https://github.com/frappe/erpnext/pull/48273
There are two options :
1. Change it to State/Province to be in line with Frappe/ERPNext
2. Add context to it

pateljannat:
@mahsem I prefer option 1. Can you make the change?

mahsem:
@pateljannat Will close this and make new PR

PR #1619, opened 2025-07-03
fix: ensure tags wrap correctly to prevent overflow and breaking of layout

JoeBrar:
I noticed that while creating a new course, if you add multiple tags the tags would overflow out of the container. And you could not click the close (x) button on these overflowing tags. (see image below)

Screenshot 2025-07-03 231159

I have updated it so the tags wrap neatly and don't break the layout.

Screenshot 2025-07-03 234315

### DIFF
--- a/frontend/src/pages/CourseForm.vue
+++ b/frontend/src/pages/CourseForm.vue
@@ -67,7 +67,11 @@
 									<FormControl
 										v-model="newTag"
 										:placeholder="__('Add a keyword and then press enter')"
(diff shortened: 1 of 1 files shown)


pateljannat:
@JoeBrar please install pre-commit to ensure that Linters GA passes

JoeBrar:
> @JoeBrar please install pre-commit to ensure that Linters GA passes

@pateljannat I have installed pre-commit and fixed the linter (prettier) issues. Please see

PR #1621, opened 2025-07-03
fix: Changed image extension validation to MIME type validation

JoeBrar:
While creating a new course, batch or job, I was not able to upload certain photos whenever an image upload was required. This was because the image validation function checks for the file extension and only allows 3 extensions - jpg, jpeg, png. This is quite restrictive and does not allow for many other image types such as svg, tiff, bmp, webp, avif etc. 

Also, file extensions can be spoofed so it is not very secure. A better way would be to check for MIME type. So I have changed the validation to MIME type validation which is more secure and incorporates all types of image files.

Also made a UI change so the user can now click on empty image card to open the Image selector dialog box.

https://github.com/user-attachments/assets/a90148fd-89ac-4d9f-a459-e69a0f5c275d

### DIFF
--- a/frontend/src/pages/BatchForm.vue
+++ b/frontend/src/pages/BatchForm.vue
@@ -209,7 +209,7 @@
 									v-slot="{ file, progress, uploading, openFileSelector }"
 								>
 									<div class="flex items-center">
-										<div class="border rounded-md w-fit py-5 px-20">
+										<div class="border rounded-md w-fit py-5 px-20 cursor-pointer" @click="openFileSelector">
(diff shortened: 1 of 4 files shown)


pateljannat:
@JoeBrar, this is a good improvement to the app. With your PR, I noticed that the validateFile function has been duplicated in three places. Can you please move the function to utils and import it from there for use?

JoeBrar:
> @JoeBrar, this is a good improvement to the app. With your PR, I noticed that the validateFile function has been duplicated in three places. Can you please move the function to utils and import it from there for use?

@pateljannat Done, please see.

PR #1651, opened 2025-07-23
fix: vimeo video embed with plyr

pateljannat:
Vimeo videos had stopped working because a recent change was considering all embeds as YouTube. Removed the line of code that was setting `data-plyr-provider` as `youtube` for all sources.


pateljannat:
Tests failing because of https://github.com/stylus/stylus/issues/2938

PR #1659, opened 2025-07-27
feat(coupon-codes): Add Coupon Code feature to LMS

OsafAliSayed:
#### TODO:
- [X] Backend Structure for coupon code in lms
- [ ] Test Backend structure so that with different edge cases.
- [ ] Frontend for supporting the Coupon Code


lms/lms/doctype/lms_coupon/lms_coupon.json
```diff
@@ -0,0 +1,158 @@
...
+  },
+  {
+   "fieldname": "course",
+   "fieldtype": "Link",
+   "in_list_view": 1,
+   "label": "Course",
+   "options": "LMS Course",
+   "reqd": 1
+  },
+  {
```
NagariaHussain: User should be able to select does it apply to a batch or course, then select a course/batch if it only applies to that course/batch @OsafAliSayed

OsafAliSayed: oh right🤔. I did not take batches into consideration. will adjust it in the next commit

lms/lms/doctype/lms_coupon/lms_coupon.json
```diff
@@ -0,0 +1,177 @@
...
+   "options": "LMS Course"
+  },
+  {
+   "depends_on": "eval:doc.applicable_to=='Batch'",
+   "fieldname": "batch",
+   "fieldtype": "Link",
+   "in_list_view": 1,
+   "label": "Batch",
+   "options": "LMS Batch"
+  },
```
NagariaHussain: Use Dynamic Link instead of conditionally showing/hiding different Link fields.

NagariaHussain (changes requested):

lms/lms/doctype/lms_coupon/test_lms_coupon.py
```diff
@@ -0,0 +1,262 @@
+import frappe
+import unittest
+from datetime import date, timedelta
```
NagariaHussain: Use Frappe's utilities instead:
`today()`, etc. 

Check utils/data file in Frappe app to learn about what is available.

lms/lms/doctype/lms_coupon/lms_coupon.py
```diff
@@ -0,0 +1,65 @@
...
+		"""Increment usage count"""
+		self.used_count = (self.used_count or 0) + 1
+		self.save(ignore_permissions=True)
+
+	def get_applicable_reference(self):
+		"""Get the reference document (course or batch)"""
+		if self.applicable_to == "Course":
+			return self.course
+		elif self.applicable_to == "Batch":
+			return self.batch
```
NagariaHussain: This will also simplify once you switch to Dynamic link.

lms/lms/doctype/lms_coupon/test_lms_coupon.py
```diff
@@ -0,0 +1,262 @@
...
+				"discount_type": "Percentage",
+				"discount_value": 20,
+				"is_active": 1,
+				"max_uses": 100,
+			}
+		).insert(ignore_permissions=True)
+
+		self.assertEqual(coupon.applicable_to, "Batch")
+		self.assertEqual(coupon.batch, self.batch.name)
+		self.assertIsNone(coupon.course)
```
NagariaHussain: Testing whether frappe correctly sets these fields is not our test to write. We should test our logic like whether it throws on trying to use an invalid coupon.

lms/lms/doctype/lms_coupon/test_lms_coupon.py
```diff
@@ -0,0 +1,262 @@
...
+				"coupon_code": "BATCHREF",
+				"applicable_to": "Batch",
+				"batch": self.batch.name,
+				"discount_type": "Percentage",
+				"discount_value": 15,
+				"is_active": 1,
+			}
+		).insert(ignore_permissions=True)
+
+		self.assertEqual(batch_coupon.get_applicable_reference(), self.batch.name)
```
NagariaHussain: We should test weather a course coupon can be applied to batch or not. This is an implementation detail and we should not test implementation but interface (what is the unit of code supposed to provide to the outside world).

lms/lms/payments.py
```diff
@@ -28,14 +77,53 @@ def get_payment_link(
...
+		coupon_result = validate_coupon(coupon_code, doctype, docname, amount)
+		if coupon_result.get("valid"):
+			discount_amount = coupon_result["discount_amount"]
+			amount = coupon_result["discounted_amount"]
+			coupon_name = coupon_result["coupon_name"]
+
+			if original_total_amount > original_amount:
+				gst_amount = original_total_amount - original_amount
+				gst_percentage = gst_amount / original_amount if original_amount > 0 else 0
+				total_amount = amount + (amount * gst_percentage)
```
NagariaHussain: Can we simplify this? Too much nesting makes the code a bit unreadable.

lms/lms/doctype/lms_coupon/lms_coupon.py
```diff
@@ -0,0 +1,65 @@
...
+			return False, frappe._("Coupon is not yet valid")
+		if self.valid_till and today > self.valid_till:
+			return False, frappe._("Coupon has expired")
+
+		if self.max_uses > 0 and self.used_count >= self.max_uses:
+			return False, frappe._("Coupon usage limit exceeded")
+
+		return True, frappe._("Coupon is valid")
+
+	def calculate_discount(self, amount):
```
NagariaHussain: This is very generic. This function should just take reference DocType and reference docname and return the discount (since amount details are available there). Also, apply any validations.

Issue #1649, opened 2025-07-22
Batch Announcement to have BCC Option


harshpwctech:
While making announcements from Batches, the email sent to the students have all the email IDs of all the other students. This makes a data privacy issue for the students/users.

We can send them as BCC by making reply-to mandatory and using it as recipients as well.

PR #1667, opened 2025-08-01
refactor: Announcement mail being sent to students in BCC

harshpwctech:
While making announcements, the mail IDs of the students were being sent in "to", this was a privacy issue also leads to spamming in case a student replies to all.

This PR shall keep the students in BCC and the Reply To is now mandatory for "to"


harshpwctech:
Solved

https://github.com/frappe/lms/issues/1649

Issue #1521, opened 2025-05-21
Enhance Learning Path


michellealva:
To automatically assign students based on certain criteria to certain courses. For example, if I am using Learning to train my employees, I will design a learning path based on each dept. Each learning path will have certain courses. Once I assign a learning path to an employee they will have to complete those courses.

This saves me time from assigning individual courses to each employee.

sidharthshah:
Why not create a "Batch" per department and assign courses at "Batch" level as opposed to Individual Employee level?

pateljannat:
@sidharthshah Batches are instructor-led. Whereas courses as self learned. But there can be multiple courses on one single subject. Learning Path is meant to show students a path in which they can take these related courses. 

Example: Let's say a learning platform has 3 courses on Maths. Now, a student may come and see the 3 courses, but may not know the right order to take them. Learning path is meant to explicitly state that order.

PR #1686, opened 2025-08-14
refactor: learning path

pateljannat:
Learning Path has been refactored. Previously, there was a setting to enable Programs; if enabled, Programs was shown to students as a replacement for Courses. Additionally, by default, the order of courses was enforced for all programs.

This has now changed. When creating programs, instructors can configure whether to enforce the order or not. 
Also, they can now configure whether Programs are published. Published programs will be visible to all students. They can then enroll in the program and proceed with the courses.

https://github.com/user-attachments/assets/13db87ea-9e44-4491-aad8-bb9720c5e8d3

Screenshot 2025-08-19 at 4 04 37 PM

However, if an instructor prefers not to keep the Program public but rather manually enroll certain selected students, they can do so as well. They can also see a Progress Summary of all programs.

https://github.com/user-attachments/assets/5a5714ce-901b-4557-a6bd-10a94e9cf22c


pateljannat:
#1521

NagariaHussain:
Finally! 🚀

PR #1773, opened 2025-10-12
fix: beautifulsoap4 version upgrade to match with Frappe Framework

HUMENTH:


ankush:
@pateljannat you'll just have to remove this and make LMS's usage compatible with all versions used by Framework. You can't change Framework's dependency from LMS.

rehanrehman389:
image

@pateljannat: Getting this error message when setting up the LMS on version 15 bench.

This needs to be reverted from the main branch.

pateljannat:
@rehanrehman389 can you confirm if the issue still exists after https://github.com/frappe/lms/pull/1913/changes/1210a6aa872a23ba2188f020d51bf9b9f891758a

rehanrehman389:
yes it is fixed

PR #1777, opened 2025-10-13
feat: coupon code discount

JoeBrar:
Moderators can create and manage Coupon Codes from the Settings modal. Coupon code can either deduct a percentage from the amount or deduct a fixed value from the amount. A course or a batch must be linked to the coupon code.

https://github.com/user-attachments/assets/e5832093-2723-402d-b54f-c4afbec2de94

A student can apply coupon code on the billing page. When applied, deductions will be shown and the reduced total gets passed to the payment gateway.

https://github.com/user-attachments/assets/723c4443-2aa5-4999-b58c-e26eee5ae034

### DIFF
--- a/lms/lms/utils.py
+++ b/lms/lms/utils.py
@@ -1932,61 +1932,79 @@ def publish_notifications(doc, method):
 
 
 def update_payment_record(doctype, docname):
-	request = frappe.get_all(
+	request = get_integration_requests(doctype, docname)
+
+	if len(request):
+		data = request[0].data
+		data = frappe._dict(json.loads(data))
+		payment_doc = get_payment_doc(data.payment)
+
+		update_payment_details(data)
+		update_coupon_redemption(payment_doc)
+
+		if payment_doc.payment_for_certificate:
+			update_certificate_purchase(docname, data.payment)
+		elif doctype == "LMS Course":
+			enroll_in_course(docname, data.payment)
+		else:
+			enroll_in_batch(docname, data.payment)
+
+
+def get_integration_requests(doctype, docname):
+	return frappe.get_all(
 		"Integration Request",
 		{
 			"reference_doctype": doctype,
 			"reference_docname": docname,
 			"owner": frappe.session.user,
 		},
+		["data"],
 		order_by="creation desc",
 		limit=1,
 	)
 
-	if len(request):
-		data = frappe.db.get_value("Integration Request", request[0].name, "data")
-		data = frappe._dict(json.loads(data))
 
-		payment_gateway = data.get("payment_gateway")
-		if payment_gateway == "Razorpay":
-			payment_id = "razorpay_payment_id"
-		elif "Stripe" in payment_gateway:
-			payment_id = "stripe_token_id"
-		else:
-			payment_id = "order_id"
+def get_payment_doc(payment_name):
+	return frappe.db.get_value(
+		"LMS Payment", payment_name, ["name", "coupon", "payment_for_certificate"], as_dict=True
+	)
+
+
+def update_payment_details(data):
+	payment_id = get_payment_id(data)
+
+	frappe.db.set_value(
+		"LMS Payment",
+		data.payment,
+		{
+			"payment_received": 1,
+			"payment_id": data.get(payment_id),
+			"order_id": data.get("order_id"),
+		},
+	)
+
+
+def get_payment_id(data):
+	payment_gateway = data.get("payment_gateway")
+	if payment_gateway == "Razorpay":
+		payment_id = "razorpay_payment_id"
+	elif "Stripe" in payment_gateway:
+		payment_id = "stripe_token_id"
+	else:
+		payment_id = "order_id"
+	return payment_id
+
+
+def update_coupon_redemption(payment_doc):
+	if payment_doc.coupon:
+		redemption_count = frappe.db.get_value("LMS Coupon", payment_doc.coupon, "redemption_count") or 0
 
 		frappe.db.set_value(
-			"LMS Payment",
-			data.payment,
-			{
-				"payment_received": 1,
-				"payment_id": data.get(payment_id),
-				"order_id": data.get("order_id"),
-			},
+			"LMS Coupon",
+			payment_doc.coupon,
+			"redemption_count",
+			redemption_count + 1,
 		)
-
-		# Increment coupon usage if applicable
-		coupon = frappe.db.get_value("LMS Payment", data.payment, "coupon")
-		if coupon:
-			frappe.db.sql(
-				"""
-				UPDATE `tabLMS Coupon`
-				SET times_redeemed = COALESCE(times_redeemed, 0) + 1
-				WHERE name = %s
-				""",
-				(coupon,),
-			)
-		payment_for_certificate = frappe.db.get_value("LMS Payment", data.payment, "payment_for_certificate")
-
-		try:
-			if payment_for_certificate:
-				update_certificate_purchase(docname, data.payment)
-			elif doctype == "LMS Course":
-				enroll_in_course(docname, data.payment)
-			else:
-				enroll_in_batch(docname, data.payment)
-		except Exception as e:
-			frappe.log_error(frappe.get_traceback(), _("Enrollment Failed, {0}").format(e))
 
 
 def enroll_in_course(course, payment_name):
--- a/frontend/src/components/Settings/Coupons/CouponItems.vue
+++ b/frontend/src/components/Settings/Coupons/CouponItems.vue
@@ -120,30 +120,6 @@ watch(
 
 const saveItems = (parent = null) => {
 	return rows.value
-	/* for (const row of rows.value) {
-        if (row.name) {
-            await applicableItems.setValue.submit({
-                ...row,
-            }, {
-                onSuccess() {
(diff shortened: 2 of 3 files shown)


frontend/src/components/Settings/CouponDetails.vue
```diff
@@ -0,0 +1,230 @@
...
+					minlength="6"
+					@beforeinput="handleCodeInput"
+					@input="doc.code = $event.target.value.toUpperCase()"
+				/>
+				<FormControl
+					v-model="doc.discount_type"
+					:label="__('Discount Type')"
+					:required="true"
+					type="select"
+					:options="['Percent', 'Amount']"
```
pateljannat: Lets rename these to Percentage and Fixed Amount to add more clarity

frontend/src/components/Settings/CouponDetails.vue
```diff
@@ -0,0 +1,230 @@
...
+					type="date"
+					:description="__('Leave blank for no expiry')"
+				/>
+				<FormControl
+					v-model="doc.usage_limit"
+					:label="__('Usage Limit')"
+					type="number"
+					:placeholder="__('Unlimited')"
+				/>
+				<Switch v-model="doc.active" :label="__('Active')" />
```
pateljannat: Enabled is the term used in other features. Lets rename this to enabled. Both the label and the fieldname

frontend/src/components/Settings/CouponDetails.vue
```diff
@@ -0,0 +1,230 @@
...
+				<div class="col-span-2">
+					<div class="text-md font-medium text-ink-gray-7 mb-1 mt-2">
```
pateljannat: Can we change this to a table format? You can refer to how test cases are added in Programming Exercises.

frontend/src/components/Settings/CouponDetails.vue
```diff
@@ -0,0 +1,230 @@
...
+	expires_on: null,
+	usage_limit: null,
+	applicable_items: [],
+})
+
+const dialogTitle = computed(() =>
+	props.couponId === 'new' ? __('New Coupon') : __('Edit Coupon')
+)
+
+const getDoc = createResource({
```
pateljannat: Ideally, couponCodes list should be fetched in Coupons.vue using createListResource and this resource should be passed as model to CouponDetails.vue. Check Badges.vue for reference.

This resource can then be used to add new coupons and edit existing coupons without having to create another resource.

frontend/src/components/Settings/CouponDetails.vue
```diff
@@ -0,0 +1,230 @@
...
+const insertDoc = createResource({
+	url: 'frappe.client.insert',
+	makeParams(values) {
+		return { doc: { doctype: 'LMS Coupon', ...doc.value } }
+	},
+})
+
+function handleCodeInput(event) {
+	if (event.data && !/^[A-Za-z0-9]*$/.test(event.data)) {
+		event.preventDefault()
```
pateljannat: If this code checks that only alphabets and numbers are used, then this is not the right approach. As a user, I will add a special character, and this would make it disappear, without any feedback, and I would assume the system is malfunctioning. 

Instead, we can add a validation on save that only letters and numbers are accepted. This way, the user gets feedback

frontend/src/components/Settings/Coupons.vue
```diff
@@ -0,0 +1,144 @@
...
+				<template #prefix>
+					<Plus class="h-3 w-3 stroke-1.5" />
+				</template>
+				{{ __('New') }}
+			</Button>
+		</div>
+
+		<div v-if="coupons.data?.length" class="overflow-y-scroll">
+			<table class="w-full text-sm">
+				<thead>
```
pateljannat: Use the Frappe UI List View component. Check Badges for usage.

frontend/src/components/Settings/Coupons.vue
```diff
@@ -0,0 +1,144 @@
...
+						<td class="p-2">{{ row.expires_on || '-' }}</td>
+						<td class="p-2">
+							{{ row.times_redeemed }}/{{ row.usage_limit || '∞' }}
+						</td>
+						<td class="p-2">
+							<Badge v-if="row.active" theme="green">{{ __('Enabled') }}</Badge>
+							<Badge v-else theme="gray">{{ __('Disabled') }}</Badge>
+						</td>
+						<td class="p-2 text-right" @click.stop>
+							<Button variant="ghost" @click="confirmDelete(row)">
```
pateljannat: With the Frappe UI list view component, multiple rows can be deleted at once, too. Check ZoomSettings.vue for reference.

frontend/src/components/Settings/TransactionDetails.vue
```diff
@@ -115,10 +148,24 @@ const props = defineProps<{
 	transaction: { [key: string]: any } | null
 }>()
 
+const paymentDoc = createResource({
+	url: 'frappe.client.get',
+	makeParams() {
+		return { doctype: 'LMS Payment', name: props.transaction?.name }
```
pateljannat: We don't need this. Why did you feel the need to make this resource?

frontend/src/components/Settings/TransactionDetails.vue
```diff
@@ -115,10 +148,24 @@ const props = defineProps<{
...
+})
+
 watch(
 	() => props.transaction,
 	(newVal) => {
-		transactionData.value = newVal ? { ...newVal } : null
+		if (newVal?.name) {
+			paymentDoc.submit()
+		} else {
+			transactionData.value = null
```
pateljannat: Again, this change is not needed.

frontend/src/pages/Billing.vue
```diff
@@ -56,6 +63,35 @@
...
+								@click="applyCouponCode"
+								variant="outline"
+								>{{ __('Apply') }}</Button
+							>
+							<Button
+								v-if="appliedCoupon"
+								@click="removeCoupon"
+								variant="subtle"
+								class="bg-red-200"
+								><X class="h-4.5 w-4.5"
```
pateljannat: Improve the formatting everywhere. Install pre-commit.

frontend/src/pages/Billing.vue
```diff
@@ -56,6 +63,35 @@
 							{{ orderSummary.data.total_amount_formatted }}
 						</div>
 					</div>
+					<div class="mb-5">
+						<div
```
pateljannat: I feel this entire section should be below the transaction details, in a separate box. The current UI does not look good.

lms/lms/doctype/lms_coupon/lms_coupon.json
```diff
@@ -0,0 +1,166 @@
...
+   "fieldtype": "Int",
+   "label": "Amount Off",
+   "mandatory_depends_on": "eval:doc.discount_type=='Amount'"
+  },
+  {
+   "default": "1",
+   "fieldname": "active",
+   "fieldtype": "Check",
+   "in_list_view": 1,
+   "label": "Active"
```
pateljannat: Make it Enabled

frontend/src/components/Settings/Coupons.vue
```diff
@@ -0,0 +1,144 @@
...
+						<td class="p-2">{{ row.discount_type }}</td>
+						<td class="p-2">
+							<span v-if="row.discount_type === 'Percent'"
+								>{{ row.percent_off }}%</span
+							>
+							<span v-else>{{ row.amount_off }}</span>
+						</td>
+						<td class="p-2">{{ row.expires_on || '-' }}</td>
+						<td class="p-2">
+							{{ row.times_redeemed }}/{{ row.usage_limit || '∞' }}
```
pateljannat: I think we should show this as 2 separate columns next to each other

lms/lms/doctype/lms_payment/lms_payment.json
```diff
@@ -18,15 +18,19 @@
   "payment_details_section",
   "currency",
   "amount",
-  "amount_with_gst",
+  "coupon",
```
pateljannat: Why are there 2 fields, coupon and coupon code, and if one is the name of the coupon code document and the other is the actual coupon code applied, then why are they not near each other?

lms/lms/payments.py
```diff
@@ -23,23 +23,43 @@ def get_payment_link(
...
 	payment_gateway = get_payment_gateway()
 	address = frappe._dict(address)
-	amount_with_gst = total_amount if total_amount != amount else 0
+
+	coupon_context = None
+	if doctype in ["LMS Course", "LMS Batch"] and coupon_code:
+		try:
+			from lms.lms.utils import apply_coupon
+
+			coupon_context = apply_coupon(doctype, docname, coupon_code)
```
pateljannat: Move this piece to a separate function

lms/lms/utils.py
```diff
@@ -1867,6 +1934,18 @@ def update_payment_record(doctype, docname):
 				"order_id": data.get("order_id"),
 			},
 		)
+
+		# Increment coupon usage if applicable
+		coupon = frappe.db.get_value("LMS Payment", data.payment, "coupon")
+		if coupon:
+			frappe.db.sql(
```
pateljannat: Use Frappe ORM

PR #1790, opened 2025-10-18
fix: diff_lms_batch_from_erpnext_batch

mahsem:
rename Batch to LMS Batch to diff from erpnext (ability to translate to other languages)

### DIFF
--- a/frontend/src/components/Modals/Event.vue
+++ b/frontend/src/components/Modals/Event.vue
@@ -29,7 +29,7 @@
 								</span>
 							</div>
 						</Tooltip>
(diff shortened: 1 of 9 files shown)


pateljannat:
@mahsem The doctype is already named LMS Batch. But we don't want to show LMS Batch on the frontend too, as it's not user-friendly. Also, the context of the page is enough to differentiate it from ERPNext.

mahsem:
@pateljannat yes but Batch in LMS has different meaning from Batch in Erpnext and will overwrite translation for erpnext
Same case as in [State](https://github.com/frappe/lms/pull/1614)
Add context ?

PR #1797, opened 2025-10-21
feat: implement quiz navigation with answer persistence

### LINKED ISSUE #1333: Allow User to Go Back to Previous Question in Quiz
For quizzes (Especially timed quizzes), it will be useful to have an option to go to previous questions so that users can edit previously submitted answers. This will allow the user to skim through and attempt easier problems first and then attempt tougher problems later if there is time left.

Image

The current architecture calls an API to check if the answer is correct after every question, hence it will not be compatible if we allow users to go back, since users can check network calls to figure out the answers.

@pateljannat 
Need your opinion and ideas regarding this feature.

Kunal-debug-png:
Fixes #1333 
@pateljannat 
Add previous/next navigation buttons and paginated question jump controls allowing users to navigate freely through quiz questions. Answers are automatically saved and restored when switching questions. Navigation is hidden in show_answers mode to ensure correct answers remain secure and are validated only server-side.

### DIFF
--- a/frontend/src/components/Quiz.vue
+++ b/frontend/src/components/Quiz.vue
@@ -211,111 +211,111 @@
 						/>
 					</div>
 					<div
-					class="flex flex-col lg:flex-row lg:items-center lg:justify-between mt-4 gap-4 text-center lg:text-left"
+						class="flex flex-col lg:flex-row lg:items-center lg:justify-between mt-4 gap-4 text-center lg:text-left"
 					>
-					<div>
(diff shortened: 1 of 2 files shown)


Kunal-debug-png:
Note : The "Next" button was removed from the <span> wrapper because dynamically adding or removing questions without a fixed container size caused DOM reallocation, leading to inconsistent UI behavior.

Kunal-debug-png:
one
show_answers=true
two

PR #1814, opened 2025-10-29
fix: use file_url instead of file_name

rehanrehman389:


rehanrehman389:
@pateljannat Hi, I’m assuming all the resumes that were uploaded are public.

pateljannat:
@rehanrehman389 make them private through the same patch, and also make sure they are linked to the LMS Job Application Document. Also, the document should then be shared with the job poster so they have access to the resume

rehanrehman389:
> Also, the document should then be shared with the job poster so they have access to the resume

@pateljannat Currently, the job poster acts as a Website User, so we cannot directly share the document with them.

image

rehanrehman389:
> Also, the document should then be shared with the job poster so they have access to the resume

Currently, the LMS Student role has read access to the LMS Job Application, so everyone with this role can access the resume.
So, I changed the "Only If Creator" checkbox.

PR #1816, opened 2025-10-31
improve: robust parsing of scorm_package (string to dict) in upsert_chapter

anupamvs:
- Add JSON parsing for string scorm_package inputs
- Try frappe.parse_json() first, fallback to json.loads()
- Improves robustness when scorm_package is passed as string


pateljannat:
Don't think this functionality is breaking currently. Closing this PR

PR #1866, opened 2025-11-28
fix: open dialog directly to create quiz and assignment

muhamiyan:
Bismillah, should be to create a new quiz or new assignment via the add button.

Feature of [**Quiz**](https://github.com/frappe/lms/pull/944) and [**Assignment**](https://github.com/frappe/lms/pull/1223) in courses provides the API endpoints [/lms/quizzes/new](https://github.com/pateljannat/lms/blob/d18a6f6e73fe898fda9bdf6f228fdbbfd162f692/frontend/src/pages/QuizCreation.vue) and [/lms/assignments/new](https://github.com/pateljannat/lms/blob/eeb7fb1f78f5b76b2063c41f51d9147fdcf288c0/frontend/src/pages/AssignmentForm.vue) so it worked fine, but this **no longer works** on the latest version of the LMS.


pateljannat:
@muhamiyan This fix redirects the users to the list, but since the intention is to create a new quiz or assignment, they still need to click on the Create button. Can we send a query param from the redirectToForm function, such that if this parameter is present then we show the creation dialog directly?

muhamiyan:
thanks @pateljannat, showing the creation dialog is better, I will fix it.

muhamiyan:
Please to review, thanks @pateljannat

frontend/src/pages/QuizForm.vue
```diff
@@ -39,11 +39,37 @@
...
+		v-model="showNewQuizModal"
+		:options="{
+			title: __('Create a Quiz'),
+			size: 'sm',
+			actions: [
+				{
+					label: __('Save'),
+					variant: 'solid',
+					onClick({ close }) {
+						saveQuizCreation(close)
```
pateljannat: We don't need this here, just opening the modal from the list is enough

pateljannat (changes requested):

frontend/src/pages/QuizForm.vue
muhamiyan: Ok, thanks @pateljannat

PR #1895, opened 2025-12-10
feat: Full Text Search

pateljannat:
https://github.com/user-attachments/assets/b5e58418-8bce-412b-9382-d4698794b5b3


rehanrehman389:
@pateljannat, the video is not proper, can you re-upload if possible?

Issue #1896, opened 2025-12-10
Unable to create LMS Certificate


namanagrawaldev:
**Describe the bug**

Unable to create LMS Certificate

**To Reproduce**

1. Go to LMS Certificate and create one certificate, it throws error while creating certificate.

**Expected behavior**
It should create certificate

**Screenshots**

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/69a071f1-1a08-4991-ace1-c8c97c9bd897" />

namanagrawaldev:
@pateljannat its on latest main branch

KerollesFathy:
Hi @namanagrawaldev, what about frappe version?

namanagrawaldev:
frappe is on latest version with commit id 937f38d

<img width="1743" height="383" alt="Image" src="https://github.com/user-attachments/assets/0c3e288d-9110-42d2-8325-d5baeb7ad5e6" />

rehanrehman389:
@anupam-singh-rajput The main branch is compatible with Frappe’s version-15 branch, and frappe.in_test is not supported in version 15. It is supported from version 16 onwards. 
https://github.com/frappe/frappe/issues/32949

pateljannat:
This has been fixed with https://github.com/frappe/lms/pull/1897

PR #1901, opened 2025-12-11
fix: Use frappe.flags.in_test instead of frappe.in_test

anupam-singh-rajput:
Closes https://github.com/frappe/lms/issues/1896


pateljannat:
The issue has been fixed with https://github.com/frappe/lms/pull/1897/

Issue #1904, opened 2025-12-12
Dropdown options not visible in the certification list page


NagariaHussain:
<img width="2176" height="1576" alt="Image" src="https://github.com/user-attachments/assets/078b5ee1-2477-44e9-bae9-8fdfc03ef7f0" />

PR #1907, opened 2025-12-13
fix: Category dropdown options not visible in the certification list page

KerollesFathy:
Fixes: #1904

https://github.com/user-attachments/assets/c2b80df6-5beb-40ed-97f5-ba85924785b2

### DIFF
--- a/frontend/src/pages/CertifiedParticipants.vue
+++ b/frontend/src/pages/CertifiedParticipants.vue
@@ -33,7 +33,7 @@
 						v-model="currentCategory"
 						:options="categories.data"
 						:placeholder="__('Category')"
-						@change="updateParticipants()"
+						@update:modelValue="updateParticipants()"
 					/>
 				</div>
 			</div>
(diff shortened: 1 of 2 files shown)


frontend/src/pages/CertifiedParticipants.vue
```diff
@@ -220,6 +220,10 @@ usePageMeta(() => {
 		icon: brand.favicon,
 	}
 })
+
+watch(currentCategory, (val) => {
+	updateParticipants()
```
pateljannat: @KerollesFathy why do we need to do this because the Select field already has an @change event which called updateParticipants

lms/lms/api.py
```diff
@@ -449,8 +449,7 @@ def get_certification_categories():
 	for doc in docs:
 		category = doc.course_title if doc.course_title else doc.batch_title
 		if category not in categories:
```
pateljannat: this won't work with key value pairs of label and value, you will need to change this condition to check for duplicates

KerollesFathy: Thanks @pateljannat for the review. I’ll fix it.

frontend/src/pages/CertifiedParticipants.vue
KerollesFathy: `@change` not working 👀 , so I fixed by using `watch()`

pateljannat: @KerollesFathy  Instead of adding a watcher, replace @change with @update:modelValue="updateParticipants()"

KerollesFathy: Done ✅

Issue #1848, opened 2025-11-17
Programming Excercises Test Cases UI Bug


adelnobel:
**Describe the bug**
Adding test cases works fine but then they don't show up in the UI when trying to view them.

**To Reproduce**
See the video below, adding test cases work fine but when trying to view them later they don't show up.

https://github.com/user-attachments/assets/80cc5a58-03c3-4d08-a037-f12f77ef37a7

**Desktop (please complete the following information):**
 - MacOS
 - Chrome
 - 142.0.7444.162

PR #1910, opened 2025-12-14
fix(exercises): remove cache to fix UI test cases

royalpinto007:
fixes- https://github.com/frappe/lms/issues/1848

https://github.com/user-attachments/assets/847da13b-e094-48e2-b6bc-f1ece8a586d2


pateljannat:
@royalpinto007 removing caching is not the right solution. The problem here is reactivity.

PR #1943, opened 2025-12-24
build: support Python 3.14

rehanrehman389:
**Issue:** LMS app is affected because [Frappe](https://github.com/frappe/frappe/pull/35387) now requires Python 3.14, and some of the app’s dependencies do not yet support this version. Due to this, the app is not able to build properly.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,11 +34,11 @@ jobs:
     - name: setup python
       uses: actions/setup-python@v2
       with:
-        python-version: '3.10'
+        python-version: '3.14'
     - name: setup node
       uses: actions/setup-node@v4
       with:
-        node-version: '20'
+        node-version: '24'
         check-latest: true
     - name: setup cache for bench
       uses: actions/cache@v4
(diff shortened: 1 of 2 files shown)


rehanrehman389:
@pateljannat Should I update the CI files as well to use the latest Python and Node versions?

pateljannat:
@rehanrehman389 yes go ahead

rehanrehman389:
@pateljannat Done

pyproject.toml
```diff
@@ -4,15 +4,15 @@ authors = [
     { name = "Frappe Technologies Pvt Ltd", email = "developers@frappe.io"}
 ]
 description = "Easy to use, open source, Learning Management System"
-requires-python = ">=3.10"
+requires-python = ">=3.14"
```
pateljannat: @rehanrehman389 this can cause issues with version 14 and version 15 of framework and erpnext right? We should probably have a range here instead

rehanrehman389: @pateljannat Used the same approach followed in ERPNext

Issue #1951, opened 2025-12-29
when navigating to statistics page , dashboard is not visible


cap-code:
**Describe the bug**
when navigating to statistics page in lms applicaition , i cannot view majority of the charts there. Only chart i am able to view is course completions
**To Reproduce**
1. Go to lms application
2. Click on statistics
3. See error

**Expected behavior**
Need to view every chart visible there.
**Screenshots**

<img width="962" height="927" alt="Image" src="https://github.com/user-attachments/assets/520371ed-61c5-4d71-b34e-99c26a42eca5" />

**Additional context**
when visiting the console tab , i was able to see this error message
aceback (most recent call last):
  File "apps/frappe/frappe/app.py", line 120, in application
    response = frappe.api.handle(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/api/__init__.py", line 52, in handle
    data = endpoint(**arguments)
           ^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/api/v1.py", line 40, in handle_rpc_call
    return frappe.handler.handle()
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/handler.py", line 53, in handle
    data = execute_cmd(cmd)
           ^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/handler.py", line 86, in execute_cmd
    return frappe.call(method, **frappe.form_dict)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/__init__.py", line 1754, in call
    return fn(*args, **newargs)
           ^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/utils/typing_validations.py", line 32, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/rate_limiter.py", line 168, in wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "apps/lms/lms/lms/utils.py", line 552, in get_chart_data
    data = frappe.db.get_all(
           ^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/database/database.py", line 756, in get_all
    return frappe.get_all(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1954, opened 2025-12-30
Fix dashboard charts visibility issue

Anugrah3108:
This PR improves the clarity and usability of the chart data returned from the get_chart_data function in lms/lms/utils.py.
The update restructures the database query to use explicit aliases, making the returned data cleaner, more readable, and easier to work with in chart visualizations.

**Changes Made**

- Updated the fields parameter in the database query to return meaningful aliases:
datefield → _unit
SUM(value_field) → total
COUNT(*) → count
- Ensures more structured and predictable chart data output
- Improves developer experience and maintainability when consuming chart results

**Impact of the Issue**

 

[When navigating to statistics page , dashboard is not visible #1951](https://github.com/frappe/lms/issues/1951)

**This enhancement makes chart data easier to interpret and appear on the dashboard.**

Before:
before

After:
after

### DIFF
--- a/lms/lms/utils.py
+++ b/lms/lms/utils.py
@@ -551,7 +551,7 @@ def get_chart_data(
 
 	data = frappe.db.get_all(
 		doctype,
-		fields=[datefield, {"SUM": value_field}, {"COUNT": "*"}],
+		fields=[f"{datefield} as _unit", f"SUM({value_field}) as total", "COUNT(*) as count"],
 		filters=filters,
 		group_by=datefield,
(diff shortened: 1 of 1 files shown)


pateljannat:
@Anugrah3108 we need to make this query conditional such that it works for both v15 and v16

Anugrah3108:
@pateljannat Do v16 support SQL-style strings?

pateljannat:
@Anugrah3108 the issue has been resolved with https://github.com/frappe/lms/pull/1961

PR #1955, opened 2025-12-30
feat: enrollment request and approval workflow

Sahil1951:
Currently, when self enrollment is disabled in Frappe LMS, learners have no way to request access to a course.

This contribution adds an optional enrollment request & approval flow. Learners can request enrollment from the course page, and once approved by configured roles, they are automatically enrolled.


pateljannat:
We don't need this feature as of now. In cases where Self Learning is disabled, the admins prefer manually enrolling students to such courses.

Sahil1951:
Thanks for the clarification!

I understand the current preference for manual enrollment when self learning is disabled.
I’ll keep this feature locally for now, and we can revisit it in the future if the need arises.

Appreciate the review and feedback.

PR #1975, opened 2026-01-08
fix: update vimeo regex to support private video hashes (#1956)

### LINKED ISSUE #1956: Private Vimeo videos don't load the video player
In the edit lesson section, when I add any public video URL from Vimeo, the video player is rendered immediately. But when I try to add a private video link, it does not render.

Private videos are usually like this:
https://vimeo.com/[videoId]/[hash]

Looking forward to your help in resolving the issue.

hasnainpatil:
### Description
Fixes #1956

**The Problem:**
Private Vimeo videos use a URL format like `vimeo.com/12345/abcde` (where `abcde` is the privacy hash). The previous regex only captured the numeric ID (`12345`) and ignored the hash. Consequently, the embed player URL was missing the required `?h=hash` parameter, causing private videos to fail to load.

**The Fix:**
- Updated the Vimeo regex in `getEditorTools` to capture the optional hash segment (Group 2).
- Updated the ID extraction logic to append `?h={hash}` to the video ID if a hash is detected.

### Verification
Verified the regex logic against standard Vimeo URL formats:
- Public URL: `vimeo.com/123456` → Returns `123456`
- Private URL: `vimeo.com/123456/abcde` → Returns `123456?h=abcde`

### DIFF
--- a/lms/locale/hu.po
+++ b/lms/locale/hu.po
@@ -3,7 +3,7 @@ msgstr ""
 "Project-Id-Version: frappe\n"
 "Report-Msgid-Bugs-To: jannat@frappe.io\n"
 "POT-Creation-Date: 2025-12-19 16:05+0000\n"
-"PO-Revision-Date: 2026-01-06 14:31\n"
+"PO-Revision-Date: 2026-01-07 14:33\n"
 "Last-Translator: jannat@frappe.io\n"
 "Language-Team: Hungarian\n"
 "MIME-Version: 1.0\n"
@@ -357,7 +357,7 @@ msgstr ""
 #: lms/lms/doctype/lms_course/lms_course.json
 #: lms/lms/doctype/lms_payment/lms_payment.json
 msgid "Amount"
-msgstr ""
+msgstr "Összeg"
 
 #. Label of the amount_usd (Currency) field in DocType 'LMS Batch'
 #. Label of the amount_usd (Currency) field in DocType 'LMS Course'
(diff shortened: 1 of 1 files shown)


shxrif:
@hasnainpatil can u share a video for this pr to get approved 

u can find a private video link of vimeo on the OG issue 

also if u can add the  regex part to remove this tracking "?fl=ip&fe=ec" from vimeo url

hasnainpatil:
@shxrif Thanks for the review!

I have updated the code (commit `ef89c42`) to use the regex: 
`/(?:http[s]?:\/\/)?(?:www\.)?vimeo\.com\/(\d+)(?:\/([a-zA-Z0-9]+))?(?:\?.*)?/`

This explicitly captures the ID and Hash, while safely consuming and discarding any tracking parameters (like `?fl=ip`) so they don't dangle in the UI.

### Verification Video
Since I cannot run the full server locally, I built a **JS test harness** to verify the regex logic specifically against the tracking URLs mentioned in the issue.

https://github.com/user-attachments/assets/f50ad4b1-7924-4a02-8f47-5054e8d7cfdc



**Results shown in video:**
* **Input:** `https://vimeo.com/1148902333/7afbb00b14?fl=ip&fe=ec` (Original Problematic URL which @omagarr reported)
* **Output:** `https://player.vimeo.com/video/1148902333?h=7afbb00b14` (Clean Player URL)

The parser now handles both standard links and private links with tracking params correctly.

pateljannat:
@hasnainpatil can you please test this with both public and private videos on the actual Lesson Form in Learning. I tried at my end and private videos are not working.

Screenshot 2026-01-14 at 7 18 27 PM

shxrif:
> @hasnainpatil can you please test this with both public and private videos on the actual Lesson Form in Learning. I tried at my end and private videos are not working.
> 
> Screenshot 2026-01-14 at 7 18 27 PM

same at my end too the regex takes the private links works nicely but idk whats happening under the hood

pateljannat:
Closing due to inactivity

PR #1976, opened 2026-01-08
fix: EmptyState would not translate, returning base non translated doctype name

mrtsvr-techloop:
### Summary
When a page has no content it creates the message : No {0} (no courses, no batches, . . . )
The script is returning the doctype name without appropriate translation.

### Issue
in the page EmptyState, the props.type isn't being translated, resulting in an inconsistent phrase translation.

### Solution
Adding a translator function in the Vue page that has the job of translating the props.type.

### Before:
image

### After:
image


mrtsvr-techloop:
Issue created: #1977

pateljannat:
Its not a bug. A translation should exist for No courses as a string for this to work

PR #1997, opened 2026-01-15
fix(lesson): ensure markdown text is displayed when pasted

raizasafeel:
In content field in lessons, markdown text (from sources like StackEdit or MS Word) were pasted as empty lines, now they are pasted properly.

Behavior before:
before-fix-markdown

Behavior after fix:
after_fix


pateljannat:
@raizasafeel on pasting the heading should render like a heading. It should be converted to a heading block.

raizasafeel:
@pateljannat All of the supported styling is done on paste from markdown resources now.

PR #2012, opened 2026-01-21
fix(chapter): recalculate indices on deletion to prevent duplicate index errors

raizasafeel:
### Summary
- This PR fixes an issue where deleting a chapter created a gap in the index sequence, causing subsequent chapter creations to assign duplicate indices.

### Problem
- If chapters were indexed 1, 2, 3, 4 and Chapter 3 was deleted, the remaining ones were 1, 2, 4.
- When a user tries to add a new chapter, the new index becomes 4. So now the indices would be 1, 2, 4, 4. This causes collision.
- This duplicate index causes routing errors (in URL), breaking the course navigation.


raizasafeel:
Will add test for this after confirming the structure of tests in course deletion

PR #2030, opened 2026-01-28
Added Czech translations

limosek:
Added Czech translations for lms.


pateljannat:
Please add translations via https://crowdin.com/project/frappe

limosek:
Sorry but that web interface is not intuitive and I already have .po file. 
I tried to find how to import .po file and did not find solution. I will not edit it line by line via this web. 
Is there a way how to import it?

PR #2103, opened 2026-02-20
fix(security): remove ignore_xss_filter to enable HTML sanitization

raizasafeel:
## Summary
- Removed `ignore_xss_filter: 1` from the `body` field in Course Lesson doctype
- Enables Frappe's built-in `nh3` HTML sanitizer to protect against XSS attacks
- Affects legacy lessons which used `body` field (modern lessons use the `content` instead of `body`)

### DIFF
--- a/lms/lms/doctype/course_lesson/course_lesson.json
+++ b/lms/lms/doctype/course_lesson/course_lesson.json
@@ -160,7 +160,7 @@
  ],
  "index_web_pages_for_search": 1,
  "links": [],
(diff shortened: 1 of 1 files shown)


pateljannat:
@raizasafeel, you have changed the JSON manually, it seems. Make the change through the desk view so that the modified timestamp changes too. Migration only updates the doctypes based on modified timestamp changes

raizasafeel:
Oh yes, sorry about that! I forgot about it and ended up picking only this line with `git add -p`. It's fixed now.

PR #2107, opened 2026-02-22
fix: disabled save button on no updates

### LINKED ISSUE #2106: Save button always enabled in course settings page
**Describe the bug**
The save button on the course settings page is always enabled even though there is no change. This leads to unnecessary api calls. 

**To Reproduce**
1. Go to 'Course page'
2. Click on 'Settings Tab'
3. Don't change anything in the settings. 
4. Click 'Save' &  Observe the network tab. 

**Expected behavior**
The Save button should be enabled only when there is any change in the settings. 

**Screenshots**

Image

 - OS: Windows
 - Browser: Chrome

jagadish-7:
Fixes #2106 

**Before**
The save button always enabled even there is no change in the form. This was leading to unnecessary API calls. 

**After**
The save button only enabled when there are updates in the form. 
https://www.loom.com/share/dd4a15aff3c94813b87904bf96c35920

### DIFF
--- a/frontend/src/pages/Courses/CourseDetail.vue
+++ b/frontend/src/pages/Courses/CourseDetail.vue
@@ -13,7 +13,7 @@
 						<Trash2 class="w-4 h-4 stroke-1.5" />
 					</template>
 				</Button>
-				<Button variant="solid" @click="childRef.submitCourse()">
+				<Button :disabled="!childRef?.isDirty" variant="solid" @click="childRef.submitCourse()">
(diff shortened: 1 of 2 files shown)


pateljannat:
@jagadish-7 This has been discussed internally within the team. Although the intention is right here, this change would make the frontend forms behaviour inconsistent with the desk forms behaviour. On the desk, too, the save button is always enabled. So to keep the behaviour consistent across, we won't be merging this.

PR #2134, opened 2026-02-25
fix: add LMS prefix to batch

mahsem:
Add LMS prefix to Batch as Batch in ERPNext and LMS have different meaning
If ERPNext and LMS are installed on same instance it becomes impossible to translate as
LMS Batch (Group) overwrites ERPNext translation 

If accepted i can add prefix to rest of translation strings

### DIFF
--- a/frontend/src/components/Modals/Event.vue
+++ b/frontend/src/components/Modals/Event.vue
@@ -32,7 +32,7 @@
 								</span>
 							</div>
 						</Tooltip>
-						<Tooltip v-if="event.batch_title" :text="__('Batch')">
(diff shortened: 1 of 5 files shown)


pateljannat:
@mahsem This change will just make the Labels look ugly in learning. The context of the app is enough to prevent any confusion between both the batches.

mahsem:
> @mahsem This change will just make the Labels look ugly in learning. The context of the app is enough to prevent any confusion between both the batches.

@pateljannat Context in english is OK but in other languages it is a problem 

As in example: Swedish 

ERPNext Batch = Parti
LMS Batch = Grupp

In the end translated string in both ERNext and in LMS becomes Grupp as LMS is last to load

PR #2144, opened 2026-02-27
fix: show profile nudge as icon when sidebar is collapsed

Owaishk08:
## Issue

When the sidebar is collapsed, the profile completion nudge breaks layout
and does not follow the sidebar icon pattern.

## Fix

- Hide full profile nudge card when sidebar is collapsed
- Show icon-only version wrapped in Tooltip
- Follow existing sidebar icon spacing and styling conventions

## Result

- No layout break
- Consistent sidebar behavior
- Matches existing sidebar UX pattern

### DIFF
--- a/frontend/src/components/Sidebar/AppSidebar.vue
+++ b/frontend/src/components/Sidebar/AppSidebar.vue
@@ -91,7 +91,9 @@
 				}}
 			</div>
 			<div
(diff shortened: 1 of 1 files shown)


pateljannat:
@Owaishk08 Please install pre-commit and fix the linters

PR #2168, opened 2026-03-05
fix(persona): redirection route and skip button

raizasafeel:
## Summary
1. User lands first on Home page, so now persona redirection happens from Home.
2. Skip button in Persona was broken, now fixed.

#### Before: need to go to courses
before-continue-gif 
#### After: directly from home page
after-continue-gif 
#### Before: Does not work
before-skip-gif 
#### After: works
after-skip-gif


frontend/src/pages/PersonaForm.vue
```diff
@@ -99,13 +100,13 @@ const skipPersonaForm = () => {
 
 const roleOptions = computed(() => {
 	const options = [
-		'Trainer / Instructor',
-		'Freelancer / Consultant',
-		'HR / L&D Professional',
-		'School / University Admin',
-		'Software Developer',
+		'Instructor',
```
pateljannat: I know this was discussed during the teardown, but let's not change the options now. We already have a good number of data collected, and changing options now would mean our charts would become out of sync with the new data

pateljannat (changes requested):

frontend/src/pages/PersonaForm.vue
raizasafeel: Okay!

pateljannat:
@Mergifyio backport to main-hotfix

pateljannat:
@Mergifyio backport main-hotfix

PR #2188, opened 2026-03-12
feat: navigate between questions in quiz

### LINKED ISSUE #1333: Allow User to Go Back to Previous Question in Quiz

pateljannat:
1. Users can now navigate between questions of a quiz. 
2. This navigation will be visible only if Show Answers is disabled. This will also show up for open-ended questions.
3. When a user submits, they will see a dialog informing them of the number of questions they have attempted and the number of questions that they have left. This will ensure they get reminded before they submit.
4. If a user refreshes the page, or closes the window, the quiz will be automatically submitted.

https://github.com/user-attachments/assets/77fa3e5a-a5f5-4fec-bc6e-9753a892679c

Closes #1333

### DIFF
--- a/frontend/src/components/Quiz.vue
+++ b/frontend/src/components/Quiz.vue
@@ -16,6 +16,13 @@
 				</li>
 				<li>
 					{{
+						__(
+							'Do not refresh the page or close this window. If you do, the quiz will be submitted automatically.'
+						)
+					}}
+				</li>
+				<li>
+					{{
 						__('This quiz consists of {0} questions.').format(questions.length)
 					}}
 				</li>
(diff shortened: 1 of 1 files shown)


harshpwctech:
@pateljannat - Would recommend to add a feature of "Mark for Review" and also something like the below on the quiz page for questions to review / navigate. This will give a better UX as otherwise for lengthier quizzes going back and forth with multiple clicks may leave with bitter experience.

Screenshot 2026-03-14 at 3 52 32 PM

Issue #2045, opened 2026-02-02
UI Enhancements


pateljannat:
- [x] Batch Dashboard Improvements #2081 
- [x] Course Dashboard Improvements #2041 
- [x] Improve default certificate format #2062

## Persona
- [ ] Persona should pop up right after setup
- [ ] Better/Smaller options label in persona 

## Course Form
- [ ] Multiselect field should show selectedValues in the input
- [ ] When instructors are added to the course modal, it sticks to the border beneath.
- [ ] When " Create New " is clicked for Instructors, the settings modal opens behind it.
- [ ] Administrator and Guest should not show up in the instructor field.
- [ ] Category create new option shows create new LMS Category
- [ ] When creating new, the creation modal should not close. User loses data consistently.
- [ ] For instructor, don't show email, show full name.
- [ ] Save button should be disabled if the form is not dirty. Should show a tooltip.
- [ ] Publish should be a button
- [ ] Section for published title should not be called Settings.
- [ ] Empty state when no chapters
- [ ] Use the new date picker component for the date field
- [ ] Course description formatting goes away on save. Add a heading to see the behaviour.
- [ ] Disable Self Enrollment, should be Self Enrollment, auto checked
- [ ] Input Field padding
- [ ] Preview video description, YouTube should be capitalized.
- [ ] Related Courses should not show the current course in the list. 
- [ ] Multiselect field, when no result has a lot of padding at the bottom.
- [ ] 0 Amount validation in paid course
- [ ] Auto Install Payments app
- [ ] All checkboxes should be switches with descriptions
- [ ] If paid course, then hide paid certification
- [ ] Paid Course and Certification sections can be separate
- [ ] Certification section will have details/documentation on how to create custom templates. Should redirect them to the desk Print Format List
- [ ] Meta tags should have a description
- [ ] Unable to delete lesson as it does not unlink Notes
- [ ] Timezone should be a dropdown

## Course Page
- [ ] Show category. The category can be clickable, which opens the list view with the category filter applied
- [ ] Wrong progress summary, just started shows 101%
- [ ] Average completion rate was also wrong
- [ ] Reviews not indented properly
- [ ] Load More for reviews
- [ ] Course Dashboard, if there are many lessons, the dashboard scrolls
- [ ] Course Student Progress Modal - The lesson progress title should be sticky

## Course List
- [ ] New tab can be removed
- [ ] Live should be renamed to Published
- [ ] Search placeholder should be just Search

## Chapter Modal
[...]

PR #2196, opened 2026-03-13
Feat/category modal dialog

LeoDanielA01:
### Description
Refactored the category creation flow in Settings to use a modal dialog instead of an inline form. This improves the user experience by keeping the settings list focused and providing clear feedback during the creation process.

### Related Issues
- #2045 (Specifically the "Category creation is odd" enhancement under Settings)

### Changes
- Replaced inline category input with a `Dialog` component.
- Added a loading spinner and "Saving..." label to the "Save" button in the modal.
- Improved the UX by allowing "Enter" to submit the form.
- Cleaned up the header by removing the toggle logic for the "New" button.

### Before :

https://github.com/user-attachments/assets/320fc4ee-0ad2-4f47-b0b8-8ab5a8033f9b

### After:

https://github.com/user-attachments/assets/6adddf56-65b0-409c-b667-113ecd88ad5a

### DIFF
--- a/frontend/src/components/Settings/Categories.vue
+++ b/frontend/src/components/Settings/Categories.vue
@@ -28,8 +28,6 @@
 			</div>
 		</div>
 
-
-
 		<div class="overflow-y-scroll">
 			<div class="divide-y divide-outline-gray-modals space-y-2">
 				<div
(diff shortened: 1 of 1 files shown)


raizasafeel:
Over here, modal over another modal does not seem ideal for singular field forms, so I am closing the PR. For consistency, we will update the other category create buttons to also use inline.

PR #2197, opened 2026-03-13
fix: hide evaluators search when list is empty

LeoDanielA01:
### Description
Hides the search bar in the Evaluators settings when the list is empty.

### Related Issues
- Part of #2045 (Fixes: "Evaluator search should not appear when there is no evaluator")

### Changes
- Added a `v-if="evaluators.data?.length > 0"` condition to the search `FormControl` in [Evaluators.vue](cci:7://file://wsl.localhost/Ubuntu-22.04/home/onebook/frappe_pr/apps/lms/frontend/src/components/Settings/Evaluators.vue:0:0-0:0).

### Before

Screenshot 2026-03-13 232700

### After

Screenshot 2026-03-13 232712

### DIFF
--- a/frontend/src/components/Settings/Evaluators.vue
+++ b/frontend/src/components/Settings/Evaluators.vue
@@ -21,7 +21,7 @@
 
 		<div class="mt-8 pb-5">
 			<FormControl
-				v-if="evaluators.data?.length > 0"
+				v-if="evaluators.data?.length > 0 || search"
 				v-model="search"
 				:placeholder="__('Search')"
 				type="text"


pateljannat:
@LeoDanielA01, this will cause problems for one scenario. Consider there is a list of evaluators. Now I search for an evaluator that does not exist in the list. This will update the Evaluators query and evaluators.data.length becomes 0. This hides the searchbar. Now there are no evaluators in the list, and the search bar is also gone. So no way for me to undo the search query.

LeoDanielA01:
@pateljannat Thanks for highlighting that. I will update the functionality to ensure the search bar remains visible even when the search result is empty, so users can modify or clear the query.

PR #2241, opened 2026-03-23
fix: add overflow-y-auto to settings dialog content panel (#2079)

### LINKED ISSUE #2079: Overflow-y set to scroll in settings dialog tabs under lists section
**Describe the bug**
The vertical overflow always set to scroll, which doesn't looks good when there are items within height. 

**To Reproduce**
1. Go to 'Settings'
2. Click on 'Any of the tab under lists section'
3. The vertical scrollbar is always visible even there is only one item due to **overflow-y-scroll** property. 

**Expected behavior**
The overflow should be auto, to enable the vertical scrollbar when items exceeds the height. 

**Screenshots**

Image

 - OS: Windows 11
 - Browser Chrome

NuriaAmoros:
Fixes #2079

The settings dialog content panel was missing `overflow-y-auto`, 
causing lists to overflow without a scrollbar.

### DIFF
--- a/frontend/src/components/Settings/Settings.vue
+++ b/frontend/src/components/Settings/Settings.vue
@@ -31,7 +31,7 @@
 				<div
 					v-if="activeTab && data.doc"
 					:key="activeTab.label"
-					class="flex flex-1 flex-col p-8 bg-surface-modal overflow-x-auto"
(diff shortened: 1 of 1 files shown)


pateljannat:
This is not solving the problem highlighted in the issue

PR #2243, opened 2026-03-23
feat: enforce lesson completion before advancing to next

### LINKED ISSUE #2178: Enforce Lesson Completion
- Every course will have a setting for this.
- If enabled, lesson completion will be enforced. They won’t be able to see the next button.
- Even if they change the URL to access a lesson without completing the current lesson, they should be redirected to their current

NuriaAmoros:
Fixes #2178

Added a check in switchLesson() to prevent students from 
advancing to the next lesson if the current one is not 
completed. Shows a warning toast message instead.

### DIFF
--- a/frontend/src/pages/Lesson.vue
+++ b/frontend/src/pages/Lesson.vue
@@ -581,6 +581,10 @@ const breadcrumbs = computed(() => {
 })
 
 const switchLesson = (direction) => {
+	if (direction === 'next' && !lesson.data.progress) {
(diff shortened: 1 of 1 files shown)


pateljannat:
@NuriaAmoros This PR does not address the points mentioned in the issue. The issue mentions that there must be a setting. The next button should not be visible, and if users try to switch lessons through the url, they should be redirected to the last lesson.

Please make the right fix and then reopen the PR

PR #2246, opened 2026-03-24
fix: misc issues in batch

NagariaHussain:
## 1. Show Preview Video in Batch

There is a field for video, but it was never displayed anywhere. Now it does:

CleanShot 2026-03-24 at 15 25 21@2x

## 2. Show Error Toast on Upload failure

Before this, it used to just silently fail without any feedback.

### DIFF
--- a/frontend/src/components/CourseOutline.vue
+++ b/frontend/src/components/CourseOutline.vue
@@ -75,6 +75,12 @@
 										/>
 									</Tooltip>
 								</div>
+								<Check
+									v-if="
+										chapter.is_scorm_package && isScormChapterComplete(chapter)
+									"
+									class="h-4 w-4 text-green-700"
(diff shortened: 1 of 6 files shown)


frontend/src/pages/Batches/components/BatchOverlay.vue
```diff
@@ -1,100 +1,108 @@
...
-						{{ __('Register Now') }}
-					</span>
+					{{ __('Enroll Now') }}
 				</Button>
-			</router-link>
-			<Button
-				variant="solid"
-				class="w-full mt-2"
-				v-else-if="
-					batch.data.allow_self_enrollment &&
```
NagariaHussain, line 87: Note: These are formatting only changes by pre-commit

PR #2293, opened 2026-04-06
fix: misc issues

pateljannat:
1. Sidebar links were not properly visible in mobile view
2. When switching themes, the course card gradient used to not change, and needed a page reload to update
3. Closes #2279 
4. Closes #2287


markmarkoh:
FYI changing the variables in lms/lms/doctype/lms_certificate/lms_certificate.py broke our email templates. 

Could you help me understand how to stay ahead of non-backwards-compatible breaking changes like these?

PR #2312, opened 2026-04-14
Fix: show chapter title in lesson view

aditya-patwa18:
This PR fixes the issue where the chapter title was only visible in Zen mode.

The chapter and course title are now displayed consistently in the lesson view, making it easier for users to understand the context of the lesson.


raizasafeel:
Could you add a detailed description with before and after? Because if you mean inside the lesson editor, it is a deliberate design choice since in normal mode, you can see the chapter name in breadcrumbs.

aditya-patwa18:
Thanks for pointing that out.

What I noticed while using the lesson view is that the chapter name is technically present in the breadcrumbs, but it’s easy to overlook when you're focused on the lesson content. Also, there’s a small inconsistency — in Zen mode, the chapter name appears near the title, but in normal mode it doesn’t.

**Before:**

* Chapter name is only visible in the breadcrumbs
* Not shown near the lesson title in normal view
* Shown near the title only in Zen mode

**After:**

* Chapter name is always visible below the lesson title
* Same behavior in both normal and Zen modes

The idea was just to make the context of the lesson clearer without changing how navigation works.

raizasafeel:
Hey, in zen mode, it's simplified due to lack of other details. In normal mode, this would make things more cluttered. Since this is an intended design decision, I am closing the PR.

PR #2320, opened 2026-04-18
fix: sidebar overflow in collapsed state

Owaishk08:
## Before
1) Sidebar was visible in full width even when collapsed.

image

## After
1) When collapsed and there are no icons to scroll, the sidebar is properly hidden.

image<br>

2) When sidebar has enough icons that require scrolling, the sidebar remains visible and scrollable.

image


raizasafeel:
Can you move the classes to the static class instead of the conditional one?

raizasafeel:
Thanks, merging!

PR #2326, opened 2026-04-22
Add secrets to build workflow for apps_json

arush15june:


raizasafeel:
Hey while raising any PR in the future, please provide description with details. As for this one, I am closing it since our apps.json only contains public repos and branches (this does not add any meaningful protection).

PR #2385, opened 2026-05-12
feat: add automatic marking for User Input questions

IridescentGil:


raizasafeel:
We already do this, submit_quiz already routes User Input → check_input_answers() .

PR #2386, opened 2026-05-12
fix(quiz): quiz question fields are now correctly populated

IridescentGil:
Quiz fields were not being populated for User Input questions due to the index not being reset correctly


raizasafeel:
This has already been fixed in develop with a newer feature allowing users to add upto 10 options per quiz

PR #2387, opened 2026-05-13
fix(persona) :Adjust persona form container width to fit content

kcashique:
## Summary

The card was fixed at sm:w-96 (384px) but the FormControl select
trigger sizes itself to the longest option text via an internal
sizer span. This caused the trigger to overflow the card boundary.

Changed sm:w-96 to sm:w-fit so the card grows to accommodate the
trigger width naturally on desktop, while w-full (already present)
handles mobile viewports below the sm: breakpoint

### Before

Screenshot from 2026-05-13 15-07-46

---

### After
Screenshot from 2026-05-13 15-17-15

### DIFF
--- a/frontend/src/pages/PersonaForm.vue
+++ b/frontend/src/pages/PersonaForm.vue
@@ -10,7 +10,7 @@
 				</span>
 			</div>
 			<div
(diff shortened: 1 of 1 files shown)


raizasafeel:
The persona has been fixed completely now

PR #2409, opened 2026-05-18
fix: resolved race condition for get certificate button

### LINKED ISSUE #2408: Race condition in create_certificate allows duplicate certificates per user per course
**Describe the bug**
The create_certificate API endpoint is vulnerable to a race condition that allows duplicate certificates to be created for the same user and course. When a user clicks the "Get Certificate" button multiple times in quick succession, multiple parallel requests pass both the is_certified() check and the validate_duplicate_certificate() validation before any of them commit to the database, resulting in duplicate LMS Certificate records.

This inflates the certification count reported by get_chart_details and creates orphan certificate records.
**To Reproduce**

1. Go to a course with enable_certification enabled
2. Complete the course (progress = 100%)
3. Click the "Get Certificate" button rapidly multiple times (3-5 clicks within 1-2 seconds)
4. Check LMS Certificate list — multiple duplicate certificates are created for the same user and course

**Expected behavior**
Only one certificate should be created per user per course, regardless of how many times the button is clicked. Subsequent clicks should return the existing certificate (as is_certified() intends) rather than creating duplicates.

- OS: Ubuntu 22.04
- Browser: Chrome 136
- Version: Frappe LMS v2.54.1

cap-code:
1. Add server-side locking (SELECT ... FOR UPDATE) in create_certificate to serialize concurrent requests for the same user+course, preventing duplicate certificates from being created when the "Get Certificate" button is clicked rapidly

2. Add a composite unique index on (member, course) via on_doctype_update as a database-level safety net, with a DuplicateEntryError handler that gracefully returns the existing certificate.

3. Add a frontend loading guard and disable the "Get Certificate" button while a request is in flight to prevent unnecessary duplicate submissions.

FIXES #2408

### DIFF
--- a/frontend/src/components/CourseCardOverlay.vue
+++ b/frontend/src/components/CourseCardOverlay.vue
@@ -77,18 +77,20 @@
 						{{ __('Start Learning') }}
 					</span>
 				</Button>
-				<Button
-					v-if="canGetCertificate"
-					@click="fetchCertificate()"
-					variant="subtle"
-					class="w-full mt-2"
-					size="md"
-				>
-					<template #prefix>
-						<GraduationCap class="size-4 stroke-1.5" />
-					</template>
-					{{ __('Get Certificate') }}
-				</Button>
+			<Button
+				v-if="canGetCertificate"
+				@click="fetchCertificate()"
+				:disabled="certificate.loading"
+				:loading="certificate.loading"
+				variant="subtle"
+				class="w-full mt-2"
+				size="md"
+			>
+				<template #prefix>
+					<GraduationCap class="size-4 stroke-1.5" />
+				</template>
+				{{ __('Get Certificate') }}
+			</Button>
 			</div>
 			<div class="space-y-3">
 				<div class="font-medium text-ink-gray-9">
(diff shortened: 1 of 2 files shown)


raizasafeel:
Could you go through [contribution guidelines](https://github.com/frappe/erpnext/wiki/Contribution-Guidelines)? We want to ideally use direct sql as little as possible. Could you check out alternatives such as `frappe.db.add_unique` or simply look into frappe query builder?

For the duplicates: the real fix is a DB-level unique constraint on `member, course` (think frappe.db.add_unique()), since that's what actually stops two simultaneous request from both creating a certificate (an if exists check alone can't, because both requests pass the check before either inserts). Keep an if certificate exists, return it as a fast path, but also wrap the insert so that if the unique constraint trips on a race, you catch it and return the existing one instead of erroring.

We'll also need a data patch here, the unique constraint won't apply on sites that already have duplicate certificates. Since that one deletes data, I'm happy to add it as a co author once your part is in.

Also could you resolve the conflicts  (with develop branch) and add a simple test (insert a certificate, attempt a second for the same member+course, assert it's rejected / count stays 1)?

Let me know if you need any help with it. Thanks for the contribution.

raizasafeel:
Closing due to inactivity

PR #2438, opened 2026-06-04
fix: zoom live class not auto-recording

NagariaHussain:
## Problem

Live classes were not auto-recording even when auto recording was set to "Cloud". The Zoom Create Meeting API expects `auto_recording` (and `private_meeting`) nested inside the `settings` object, but we sent them as top-level fields — Zoom silently ignores unknown top-level fields, so the setting never took effect and meetings fell back to the host's account-level recording defaults.

## How this PR solves it

Moves `auto_recording` and `private_meeting` into the `settings` object of the meeting creation payload, as required by the Zoom API.


NagariaHussain:
ref: https://developers.zoom.us/docs/api/meetings/#tag/meetings/post/users/%7BuserId%7D/meetings

Issue #2457, opened 2026-06-12
[Bug] Student checkout fails with confusing "no role permission" toast when no payment gateway is configured (root cause in lms.lms.payments.get_payment_link)


walidsaleh:
## Describe the bug

When a student (LMS Student role) clicks "Proceder al pago" on the
billing page (`/lms/billing/<type>/<slug>`), and `LMS Settings.payment_gateway`
is empty or misconfigured, the backend raises a confusing
"PermissionError" toast that misleads the operator about the real cause.

The Spanish toast reads (translated):
> El usuario X no tiene acceso a doctype a través del permiso de rol para el documento Pasarela de Pago

In English, this suggests the student lacks a role permission for the
Payment Gateway DocType — but the real cause is that the gateway is
not configured in LMS Settings.

## Root cause

The student-facing billing page calls `lms.lms.payments.get_payment_link`
(this is correct — it's the public whitelisted endpoint meant for
students). The function (in `apps/lms/lms/lms/payments.py`, line 26)
reads `LMS Settings.payment_gateway` via `get_payment_gateway()`
(line 35), then calls `controller = get_controller(payment_gateway)`
(line 69).

`get_controller("")` in `apps/lms/lms/lms/payments.py:14` calls
`payments.utils.get_payment_gateway_controller("")` which, in
`apps/payments/payments/utils/utils.py:16`, executes
`frappe.get_doc("Payment Gateway", "")`. This raises
`frappe.DoesNotExistError` internally, but Frappe re-surfaces it as
a `PermissionError` toast when the calling user (LMS Student) has
no read permission on the Payment Gateway DocType — a defensive
behaviour to avoid leaking DocType existence to unprivileged users.

So the operator sees "no role permission" when the actual problem is
"the gateway is not configured in LMS Settings" (or it is configured
but the `gateway_controller` field points to a non-existent Settings
docname, in which case the same code path also fails at line 24 of
`utils.py`).

A second, related root cause is that `Payment Gateway` DocType has no
validation that `gateway_controller` points to a real docname of the
referenced `gateway_settings` DocType. Operators can save a Payment
Gateway with `gateway_controller="staging"` even if no
`Stripe Settings/staging` doc exists, and the error only surfaces at
checkout time, with the misleading "PermissionError" toast.

## To Reproduce

1. Fresh install of LMS via `frappe_docker` (LMS 2.55.0, Frappe 16.x,
   payments 0.0.1).
[...]

PR #2462, opened 2026-06-12
Fix lms billing razorpay hardcoded

### LINKED ISSUE #2458: [Bug] Billing.vue hardcodes Razorpay checkout script regardless of configured payment gateway
## Describe the bug

The student-facing billing page (`/lms/billing/<type>/<slug>`, served by
`apps/lms/lms/public/frontend/src/pages/Billing.vue`) injects the Razorpay
checkout script unconditionally in `onMounted`, regardless of which
payment gateway the operator has configured in `LMS Settings.payment_gateway`.

```js
// apps/lms/lms/public/frontend/src/pages/Billing.vue (Billing.vue line 75)
onMounted(() => {
  const script = document.createElement('script')
  script.src = `https://checkout.razorpay.com/v1/checkout.js`
  document.body.appendChild(script)
  if (user.data?.name) {
    access.submit()
  }
})
```

This is problematic for two reasons:

1. **It pollutes the Network tab** with a 404 red error
   (`https://checkout-static-next.razorpay.com/build/undefined`) on every
   billing page load, because the script's auto-loader fetches a
   path-derived bundle that ends up as `undefined` when no key is set.
2. **It implicitly assumes Razorpay as the default gateway**, which
   is a problem for any operator who configures a different gateway
   (Stripe, Paymob, Braintree, etc.). The `payments` app in Frappe
   supports several gateways, but this Billing.vue script injection is
   hardcoded to one specific provider.

The relevant backend code (`apps/lms/lms/lms/payments.py`) actually
DOES support arbitrary gateways via `LMS Settings.payment_gateway` →
`payments.utils.get_payment_gateway_controller()` — it's just the
frontend Billing.vue that hardcodes Razorpay.

## To Reproduce

1. Install LMS via the `frappe_docker` stack.
2. Configure `LMS Settings.payment_gateway` to "Stripe" (or any
   non-Razorpay gateway, including Paymob, Braintree, etc. from
   the `payments` app).
3. Configure `Stripe Settings` with your test API keys.
4. Open `https://<your-site>/lms/billing/course/<slug>` as a logged-in
   student.
5. Open the Network tab in DevTools.
6. Observe: a 404 red error on
   `https://checkout-static-next.razorpay.com/build/undefined` (this
   is the Razorpay script's auto-loaded bundle with an `undefined`
   path component).
7. No visible functional impact on the page itself — the script
   loads, fails with 404 silently, and the page continues. But it's
   a confusing artifact that suggests something is wrong.

## Expected behavior

The Billing.vue should only inject the Razorpay checkout script if
`LMS Settings.payment_gateway` is set to "Razorpay" (or similar).
For other gateways (Stripe, Paymob, Braintree, etc.), the frontend
should either skip the script entirely or inject the relevant
gateway's checkout SDK (e.g. `https://js.stripe.com/v3/` for Stripe).

## Additional context

This is independent of (but related to) the PermissionError bug
filed at #2457. The backend has 2-3 issues in the same checkout
flow; the frontend has at least this one Razorpay hardcoding plus
the URL [object Object] bug when the redirect fails (filed at #2459).

We have a minimal workaround in our fork (Fisiocyl) — a small
nginx sub_filter that injects a JS shim into the HTML before the
LMS bundle, which neuters `document.createElement('script')` for
Razorpay URLs. Happy to share if useful, but the real fix lives in
the upstream.

walidsaleh:
Fixes frappe/lms#2458

## What

Billing.vue's onMounted hook unconditionally injects the Razorpay
checkout script on every page load, regardless of which payment
gateway the operator has configured in LMS Settings. When the operator
uses Stripe (or any non-Razorpay gateway from the `payments` app),
this leads to:

1. A red 404 in the Network tab on
   `checkout-static-next.razorpay.com/build/undefined` (the Razorpay
   script's auto-loader fetches a path-derived bundle that becomes
   'undefined' when no API key is set).
2. The Razorpay script may interfere with the student checkout flow
   (overlapping toast/dialog), masking the actual error from the
   configured gateway.

## Why

For non-Razorpay operators (Stripe, Paymob, Braintree, etc.), every
billing page load produces a visible error in DevTools, and the
checkout UX is degraded by a third-party script that should not be
loaded.

## How

Only inject the Razorpay checkout script when
`LMS Settings.payment_gateway === 'Razorpay'`. The `useSettings`
Pinia store already loads LMS Settings on app boot, so we just read
`settings.data?.payment_gateway` in onMounted.

### DIFF
--- a/frontend/src/pages/Billing.vue
+++ b/frontend/src/pages/Billing.vue
@@ -246,6 +246,7 @@ import {
 } from 'frappe-ui'
 import { reactive, inject, onMounted, computed, ref, watch } from 'vue'
 import { sessionStore } from '../stores/session'
+import { useSettings } from '@/stores/settings'
 import Link from '@/components/Controls/Link.vue'
 import NotPermitted from '@/components/NotPermitted.vue'
 import { X } from 'lucide-vue-next'
@@ -256,11 +257,14 @@ const user = inject('$user')
 const { brand } = sessionStore()
 const showConsentWarning = ref(false)
 const { capture } = useTelemetry()
+const settings = useSettings()
 
 onMounted(() => {
-	const script = document.createElement('script')
-	script.src = `https://checkout.razorpay.com/v1/checkout.js`
-	document.body.appendChild(script)
+	if (settings.data?.payment_gateway === 'Razorpay') {
+		const script = document.createElement('script')
+		script.src = `https://checkout.razorpay.com/v1/checkout.js`
+		document.body.appendChild(script)
+	}
 	if (user.data?.name) {
 		access.submit()
 	}
--- a/lms/lms/payments.py
+++ b/lms/lms/payments.py
@@ -1,4 +1,5 @@
 import frappe
+from frappe import _
 
 from lms.lms.utils import (
 	complete_enrollment,
@@ -33,6 +34,11 @@ def get_payment_link(
 	country: str | None = None,
 ):
 	payment_gateway = get_payment_gateway()
+	if not payment_gateway:
+		frappe.throw(
+			_("No payment gateway is configured. Please contact the administrator."),
+			frappe.ValidationError,
+		)
 	address = frappe._dict(address)
 	redirect_to = get_redirect_url(doctype, docname, payment_for_certificate)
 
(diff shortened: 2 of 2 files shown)


walidsaleh:
Closing in favor of a clean PR without the cherry-picked commit of #2457's fix.
See the new PR for the same fix as a single, self-contained change.

PR #2464, opened 2026-06-12
fix(billing): guard window.location.href against non-string payload

### LINKED ISSUE #2459: Bug] Student checkout redirects to /lms/billing/course/[object Object] when backend returns error (frontend assumes string URL)
## Describe the bug

When the backend `lms.lms.payments.get_payment_link` endpoint returns
an error (e.g. because no payment gateway is configured, or the
gateway is misconfigured, or the user lacks permission), the
frontend Billing.vue takes the error response and does
`window.location.href = data` on it. But `data` is an object
like `{messages: ["..."]}` not a URL string, so the browser
navigates to a URL like:

    /lms/billing/course/[object%20Object]

instead of staying on the billing page or showing a useful error.

## To Reproduce

1. Install LMS via the `frappe_docker` stack with Stripe in test mode.
2. Either:
   - Leave `LMS Settings.payment_gateway` empty, OR
   - Point `Payment Gateway.gateway_controller` to a non-existent
     docname (e.g. `controller = "staging"` while no `Stripe Settings/staging`
     doc exists).
3. As a logged-in LMS Student, navigate to any paid course's
   `/lms/billing/course/<slug>` page.
4. Fill in the billing details form.
5. Click "Proceed to Payment".
6. Observe the browser URL: `.../lms/billing/course/[object%20Object]`.
7. The student is now on a broken page (404 from the LMS router
   because `[object Object]` isn't a valid course slug).

## Expected behavior

The student should see a useful error toast (e.g. "Payment gateway
not configured, contact the administrator") and remain on the
billing page (or be redirected back to the course page). They
should NOT be navigated to a broken URL with `[object Object]`.

## Additional context

Root cause is in `apps/lms/lms/public/frontend/src/pages/Billing.vue`,
the `onSuccess` callback of `generatePaymentLink()`:

```js
const generatePaymentLink = () => {
  paymentLink.submit(
    {},
    {
      // ...
      onSuccess(data) {
        capture('checkout_initiated', { type: props.type })
        window.location.href = data
      },
      onError(err) {
        toast.error(err.messages?.[0] || err)
      },
    }
  )
}
```

The `onSuccess` assumes `data` is a URL string (which is what
`get_payment_link` returns on success). But Frappe's resource
layer also dispatches `onSuccess` with the full response object
when the call returns a non-string (e.g. a dict error payload),
because the HTTP status was 200 (the endpoint didn't throw, it
just returned a dict). The result: the browser navigates to
`[object Object]`.

There are two fixes that work together:
- The backend `get_payment_link` should throw a proper Frappe
  exception (e.g. `frappe.throw("Payment gateway not configured")`)
  so the frontend's `onError` path runs instead of `onSuccess`.
- The frontend should defensively check `typeof data === 'string'`
  before assigning to `window.location.href`.

This is independent of (but compounded with) the PermissionError
bug at #2457 and the Razorpay hardcoding at (new issue #2). The
three together make the student checkout flow completely broken
on a fresh install with Stripe.

We have a minimal workaround in our fork (Fisiocyl) — a backend
override on `get_payment_link` that returns a redirect URL with a
query param `?error=no_payment_gateway_configured` when no gateway
is set, so the frontend redirects to a real URL (not `[object Object]`).
Happy to share if useful, but the real fix lives in the upstream.

walidsaleh:
Fixes frappe/lms#2459

## What

Billing.vue's `onSuccess` handler does `window.location.href = data`,
assuming `data` is a string URL. When the backend returns a non-string
payload (e.g. an error dict like `{messages: [...]}`), the browser
navigates to `/lms/billing/course/[object Object]` and the student
lands on a 404 page.

## Why

The student gets stuck on a broken URL with no useful error message.
The actual cause is hidden because `String({...})` produces
`[object Object]`, which gives the operator no clue about what
went wrong.

## How

Guard the assignment with a `typeof === 'string'` check. If `data`
is not a string, show a generic error toast via the existing
`toast.error` + `__()` pattern instead of navigating to a broken URL.

This is the second line of defense. The first line is the upstream
fix for `get_payment_link` (frappe/lms#2457), which now throws
`ValidationError` instead of returning a dict, so `onError` handles
the typical case. This `onSuccess` guard catches any future endpoint
that may return a non-URL payload.


raizasafeel:
The problem is same as the PR before this, seems like LLM slop.

PR #2466, opened 2026-06-12
fix(socket): use relative path so socket.io works behind any reverse …

### LINKED ISSUE #2465: [Bug] socket.io fails to connect when LMS is served behind a reverse proxy
**Describe the bug**

When the LMS frontend is served from a port that differs from
`socketio_port` (e.g. behind a reverse proxy on 443, or in dev with
`bench start` where the web is on 8000 and socketio on 9000),
`io()` fails to connect and the SPA falls back to a "no real-time
connection" state. The realtime channel (live updates, discussion
replies, etc.) silently stops working.

The root cause is in `frontend/src/socket.js`: it builds the
connection URL by concatenating `window.location.hostname`, an
optional `:${socketio_port}` and the site name, which produces an
invalid host.

**To Reproduce**

1. Deploy LMS behind nginx (or any reverse proxy) following Frappe's
   standard production setup. The vhost proxies `/` to the gunicorn
   web (8000) and `/socket.io/` to the socketio worker (9000). The
   site is served at `https://example.com`.
2. Open `https://example.com/lms` in the browser.
3. Open the browser devtools, Network tab, filter by `socket.io`
   (or `WS` for the websocket upgrade).
4. Observe: the client tries to open
   `https://example.com/<site_name>/socket.io/...` — note the
   `<site_name>` segment in the path, not the host.
5. Watch the request fail with a 404 (nginx cannot route the
   `/<site_name>/socket.io/...` prefix).

The same failure also reproduces in dev with `bench start`:
`window.location.port` is `8000` and `socketio_port` is `9000`, so
the code produces `http://localhost:8000:9000/...` which the browser
parses as host `localhost:8000:9000` and fails to resolve.

**Expected behavior**

The client should connect to the same origin that served the HTML
(`https://example.com`) at the standard `/socket.io/` path, so the
reverse proxy can route it to the socketio worker. In dev with
`bench start`, the client should connect directly to the socketio
worker on `socketio_port`.

**Screenshots**

Not applicable (network-only behaviour, best observed in devtools).

- OS: any (reproduces on macOS 14, Ubuntu 22.04, Debian 12)
- Browser: Chrome 124+, Firefox 124+, Safari 17+
- Version: develop branch of `frappe/lms`

**Smartphone (please complete the following information):**

Not tested, but the bug is platform-independent (vanilla
socket.io-client).

**Additional context**

The URL construction in `frontend/src/socket.js` produces an
invalid host in both production (the site name ends up as a path
prefix on the socket.io endpoint) and dev (the port gets mangled
into the host component). The standard `/socket.io/` path that
python-socketio serves is never reached.

walidsaleh:
Fixes frappe/lms#2465

## What

`frontend/src/socket.js` builds the socket.io connection URL by
concatenating `window.location.hostname`, an optional `:${socketio_port}`
and the site name. In production behind a reverse proxy the result
is `https://<host>/<site_name>`, where the site name ends up as a
path prefix on the socket.io endpoint. nginx returns 404 because
no location block matches `/<site_name>/socket.io/`. In dev with
`bench start` the same logic produces `http://localhost:8000:9000/...`,
which the browser parses as host `localhost:8000:9000` and fails to
resolve.

## How

Drop the manual URL construction. When the frontend is served from
a standard web port (no port, 80 or 443), use `window.location.origin`
and let the reverse proxy forward `/socket.io/` to the socketio
worker. When the frontend is served from a non-standard port (dev
`bench start`), connect directly to the socketio worker on
`socketio_port`. Set `path: '/socket.io/'` explicitly to document
the intent.

## Testing

Verified locally in a self-hosted deployment behind nginx. Before
the fix, opening `https://<host>/lms` resulted in failed WS
handshakes to `wss://<host>/<site_name>/socket.io/`. After the fix,
the client connects to `wss://<host>/socket.io/` and the realtime
channel works.

Also verified that `bench start` (web on 8000, socketio on 9000)
still works: the `isSameOrigin` check falls through to the else
branch and connects directly to the socketio worker.


raizasafeel:
Thanks for digging into this. The entire PR is kind of a mess, I am not sure why there's a diff against init.py with the project version. Also it pulls unrelated commits into the PR. Could you re-open against **develop** with just the single socket.js commit?

PR #2487, opened 2026-06-20
fix: show chapter title in lesson editor breadcrumb

### LINKED ISSUE #635: UX: show chapter title in lesson
Currently I don't know which chapter's lesson I am editing. 

Created live on #BWH

danishsshaikh:
### What changed

Adds the selected chapter title and lesson title to the Course Editor breadcrumb so instructors can tell which chapter's lesson they are editing.

The chapter breadcrumb item is intentionally not routed because chapters are outline sections, not standalone pages.

A follow-up improvement could update frappe-ui's Breadcrumbs component to render non-actionable items as plain text instead of button elements.

Closes #635 

### Screenshots

Before:

before-lesson-editor-breadcrumb

After:

after-lesson-editor-breadcrumb

### Tests

- `git diff --check`
- `cd frontend && ./node_modules/.bin/vitest run`

### DIFF
--- a/frontend/src/pages/Courses/CourseDetail.vue
+++ b/frontend/src/pages/Courses/CourseDetail.vue
@@ -182,6 +182,7 @@ interface EditorSelection {
 	chapterNumber: string
 	lessonNumber: string
 	number: string
+	chapterTitle?: string
 	title?: string
 }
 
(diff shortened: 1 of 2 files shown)


raizasafeel:
The breadcrumbs right now are a deliberate decision. If we do this instead, it wouldn't make sense for the other tabs like Settings, Dashboard and Overview. Instead, for this, I've made it so that the lesson you are on is always highlighted with the accordion of the current chapter being open by default. Thank you for the contribution but I will have to close this PR.

danishsshaikh:
Thanks for the clarification, the highlighted lesson + opening the current chapter accordion sounds like a better fit for the current product direction. Appreciate you reviewing the PR and explaining.

PR #2488, opened 2026-06-20
fix(editor): embed URLs not rendering when pasted or typed in lesson editor

DeployThemAll:
## Summary

- Pasting or typing embed URLs (YouTube, Vimeo, Bunny Stream, etc.) in the lesson editor only worked on the first paste. Subsequent pastes stayed as plain text.
- The root cause: the Embed tool requires `{ service, source, embed }` but the Markdown tool only passed `{ source }`, so the Embed tool rendered empty blocks.
- The paste handler had no embed detection path — URLs fell through to plain text.

## Changes

- Add `_resolveEmbedService()` to match pasted/typed URLs against all configured embed services and build the full embed data object
- Add `_parsePastedEmbedBlocks()` to detect embed URLs on paste; bails out entirely if any line is not a recognized embed URL (prevents swallowing mixed text+URL pastes)
- Add embed handling in `_onNativePaste` before markdown detection
- Append an empty markdown block after embeds so the user can continue typing
- Fall back to insert+delete in `_convertBlock` for tools without `conversionConfig`
- Remove unused `_isEmbed()` replaced by `_resolveEmbedService()`

## Test plan

- [ ] Paste a YouTube URL into an empty markdown block → should render as embed
- [ ] Paste a second YouTube URL below → should also render (not stay as text)
- [ ] Type a YouTube URL and press Enter → should convert to embed
- [ ] Paste a Vimeo/Bunny/CodePen URL → should render via matching service
- [ ] Paste mixed text + URL → should NOT convert (falls through to plain text)
- [ ] Paste multiple embed URLs (one per line) → all should render as embeds

### DIFF
--- a/frontend/src/utils/markdownParser.js
+++ b/frontend/src/utils/markdownParser.js
@@ -1,6 +1,6 @@
 import { CodeXml } from 'lucide-vue-next'
 import { createApp, h } from 'vue'
-import { escapeHTML } from '@/utils'
+import { escapeHTML, getEditorTools } from '@/utils'
 
 export class Markdown {
 	constructor({ data, api, readOnly, config }) {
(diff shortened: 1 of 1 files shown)


raizasafeel:
Can't recreate this issue locally with youtube link (pasted same link twice and rendered as videos), could you **first please open an issue** with bug tag recreating it then link the same here? Also please add screenshots, learning version used, etc.

**Edit:** Only half of the description is true. This bug seems to only exist for typed videos and the fix is also only for the same. Please rewrite the commit message, PR title and PR description for the same.

raizasafeel:
Closing due to inactivity

PR #2489, opened 2026-06-20
feat(editor): expose EditorJS instance on window for plugin extensibility

DeployThemAll:
## Summary

- Third-party Frappe apps that extend the LMS editor (e.g. video hosting plugins, audio material plugins) need to insert blocks programmatically via `editor.blocks.insert()`.
- In production Vue 3 builds, `setupState` is empty and `__vueParentComponent` is not set on DOM elements — the EditorJS instance created in `LessonForm.vue` is completely inaccessible from outside the component.
- This one-line change sets `window.__lmsEditor` when the editor initializes, providing a stable public API for plugins.

## Changes

- Set `window.__lmsEditor = editor.value` in `LessonForm.vue` after editor creation

## Test plan

- [ ] Open lesson editor → `window.__lmsEditor` should be the EditorJS instance
- [ ] `window.__lmsEditor.blocks.insert(...)` should work from browser console
- [ ] Existing editor functionality unchanged (save, load, toolbar, etc.)
- [ ] Navigate away from editor → `window.__lmsEditor` remains (cleaned up on next mount)

### DIFF
--- a/frontend/src/pages/LessonForm.vue
+++ b/frontend/src/pages/LessonForm.vue
@@ -89,6 +89,7 @@ onMounted(() => {
 	}
 	capture('lesson_form_opened')
 	editor.value = renderEditor('content')
+	window.__lmsEditor = editor.value
 	instructorEditor.value = renderEditor('instructor-notes')
 	window.addEventListener('keydown', keyboardShortcut)
 	enablePlyr()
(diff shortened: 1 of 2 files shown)


raizasafeel:
Closing this. The diagnosis is right i.e. the editor instance isn't reachable in a prod Vue build, **but a window.__lmsEditor global isn't the fix.** It's an undocumented, unversioned, XSS-reachable handle that core would have to support forever for one plugin's convenience.

Anyone extending the editor this deeply is already writing a custom app and can modify editor.js directly. I think that would be the right place for it, not a global escape hatch in core.

PR #2506, opened 2026-06-24
feat: Arabic translations for LMS (99.1% coverage)

medo94my:
## What
Adds full Arabic (ar) translation support for Frappe LMS.

### Changes
- **1816 Arabic translations** in `lms/translations/ar.csv` covering all LMS UI strings
- **Fix null language fallback** in `get_translations()` — when user language is null, falls back to system default

### Coverage
- 1816/1833 strings translated (99.1%)
- Remaining 17 are untranslatable JS code artifacts

### Testing
Deployed and tested on live instance. Desk sidebar, LMS frontend, onboarding, and settings all render in Arabic with proper RTL.

PR #2512, opened 2026-06-27
feat: wrap UI strings in __() and add RTL support

medo94my:
## What

Follow-up to #2506, split out as requested by @raizasafeel ("Could you raise another PR with just the rest of the fixes?"). The Arabic translation CSV is **excluded** — translations are contributed via Crowdin. This PR contains only the supporting code fixes that benefit all locales.

## Changes

- **Wrap ~95 hardcoded English strings with `__()`** across Settings, Sidebar (AppSidebar, UserDropdown), Statistics, Programming Exercises, Batches, Programs, and other components — so they become translatable in every Crowdin language, not just Arabic.
- **Fix null-language fallback in `get_translations()`** (`lms/lms/api.py`): a logged-in user with no `language` set previously passed `null` to `get_all_translations()`. It now falls back to System Settings, then `"en"`. The duplicate System Settings lookup is collapsed into a single query.
- **RTL support in `index.css`**: `html[dir='rtl']` font-family override, preserving monospace fonts for `code`/`pre` blocks — benefits any RTL locale (ar/he/fa/ur).
- **Load IBM Plex Sans Arabic** for proper RTL rendering.

## Addresses prior review

The two P1 items the automated review raised on #2506 are already handled here:
- `Settings.vue` `doctype` is a plain string (`ref('LMS Settings')`), **not** wrapped in `__()` — so the DocType lookup is unaffected.
- The translation-key whitespace issue was specific to `ar.csv`, which is not part of this PR.

## Testing

Deployed and tested on a live instance: Desk sidebar, LMS frontend, onboarding, and settings render correctly, and RTL layout applies as expected.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -201,6 +201,9 @@
 			media="(device-width: 320px) and (device-height: 568px) and (-webkit-device-pixel-ratio: 2) and (orientation: landscape)"
 		/>
 		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+		<link rel="preconnect" href="https://fonts.googleapis.com" />
+		<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+		<link href="https://fonts.googleapis.com/css2?family=IBM+Plex+Sans+Arabic:wght@300;400;500;600;700&display=swap" rel="stylesheet" />
 		<title>{{ title | e }}</title>
 		<meta name="title" content="{{ meta.title | e }}" />
 		<meta name="image" content="{{ meta.image | e }}" />
--- a/frontend/src/components/Layouts/MobileLayout.vue
+++ b/frontend/src/components/Layouts/MobileLayout.vue
@@ -121,7 +121,7 @@ const addOtherLinks = () => {
 		addLink('Profile', 'UserRound')
 		addLink('Log out', 'LogOut')
 	} else {
-		addLink('Log in', 'LogIn')
+		addLink(__('Log in'), 'LogIn')
 	}
 }
 
@@ -163,7 +163,7 @@ const addAssignments = () => {
 }
 
 const addProgrammingExercises = () => {
-	addLink('Programming Exercises', 'Code', 'ProgrammingExercises')
+	addLink(__('Programming Exercises'), 'Code', 'ProgrammingExercises')
 }
 
 const addPrograms = async () => {
--- a/frontend/src/components/Modals/LiveClassModal.vue
+++ b/frontend/src/components/Modals/LiveClassModal.vue
@@ -137,7 +137,7 @@ const getTimezoneOptions = () => {
 const getRecordingOptions = () => {
 	return [
 		{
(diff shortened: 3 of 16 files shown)


raizasafeel:
Thanks for sending this separately, could you rebase on top of latest develop (11 conflicts, and `LiveClassModal.vue`/`BulkCertificates.vue` are both deleted upstream now). Also change these:

1. `or "en"` in `get_translations()` — we already have that in the backend via `frappe.local.lang`, its not needed. Just call `get_all_translations(frappe.local.lang)`, same as `get_all_translations` already guards the null case. Also this whole function is basically reimplementing what `frappe/translate.py` already does (form dict → cookie → header → user → system settings), so we lose the guest language detection by hardcoding it.

2. `MobileLayout.vue` — `addLink(__('Log in'), 'LogIn')` breaks the button. `handleClick` compares `tab.label == 'Log in'`, so once it's translated that check fails and the click does nothing. Same as the bug in f92513789 — revert to `addLink('Log in', 'LogIn')` and translate at render (`{{ link.label }}`) instead. Same issue applies to `isVisible()` and `filterLinksToShow()`, both match/slugify on `tab.label` — check those too.

3. Split the RTL font stuff into its own PR, it's unrelated to the `__()` work and needs a different approach anyway — see 4 and 5.

4. Drop the Google Fonts links in `index.html`. Loads on every page/locale for a font only RTL uses, and it's a third-party request self-hosted/air-gapped instances shouldn't need.

5. Same for the `html[dir='rtl'] *` block in `index.css` — don't fight Tailwind's `rtl:`/logical-properties setup with a separate `!important` rule. It also breaks `font-mono` for the code submission textarea (`ProgrammingExerciseSubmission.vue:73`), which is the one place we really don't want a proportional font. Better to just add IBM Plex Sans Arabic to the `sans` stack in `tailwind.config.js` and let the browser fall back per-glyph — no `dir` check needed, and the font links go away too.

6. `AppSidebar.vue` — leave `__('Frappe Learning')` untranslated, it's a product name, same class of bug as #2 if it ever gets matched against English somewhere.

7. Prettier's failing on `Settings.vue`, `SettingFields.vue`, `ProgrammingExercises.vue` — lines are past 80 chars now. Needs to be clean before CI will pass. Please read frappe's contribution guidelines and lint your code accordingly

sashalab:
Additional confirmation for the null-language case on a self-hosted Frappe v16 deployment with social-login provisioning: an authenticated user with an unset User.language can receive an empty translation dictionary even when the site language is Russian. The current LMS main at 4b730704 still reads User.language directly, while Frappe version-16 at 5cba016e returns {} for an empty language. This is separate from missing translation strings. The proposed use of frappe.local.lang in the maintainer review would address the fallback without requiring per-user backfills. A regression test covering a logged-in user with no explicit language and a non-English system language would be valuable.

raizasafeel:
Closing this due to inactivity, please reopen with requested changes

PR #2544, opened 2026-07-05
fix: duplicate courses shown to moderators in created courses list

### LINKED ISSUE #2476: [Bug] get_created_courses() moderator fallback missing .distinct() causes duplicate course cards on Home page
**Describe the bug**
On the LMS Home page, the "Courses Created" section displays the same course multiple times when logged in as a Moderator. The number of duplicate cards equals the number of instructors added to that course.

**To Reproduce**
1. Create an LMS Course and add 3 instructors in the Course Instructor child table
2. Log in as a user who has the Moderator role but is NOT listed as an instructor in any course
3. Go to the Home page
4. The same course appears 3 times in "Courses Created" section
5. Other courses may not appear at all (hidden because the .limit(3) is exhausted by duplicates)

**Root Cause**
In `lms/lms/api.py`, the `get_created_courses()` function has a moderator fallback:

```python
if not len(results) and ("Moderator" in roles):
    results = base_query.run(as_dict=True)
```

The `base_query` does a JOIN between `CourseInstructor` and `LMS Course` without `.distinct()`. This means for a course with N instructors, the query returns N rows for that same course. Combined with `.limit(3)`, a single course with 3 instructors fills all 3 slots, preventing other courses from appearing.

**Expected behavior**
Each course should appear exactly once in the "Courses Created" section regardless of how many instructors are assigned to it.

**Suggested Fix**
Add `.distinct()` to the base_query select:

```python
base_query = (
    frappe.qb.from_(CourseInstructor)
    .join(Course)
    .on(CourseInstructor.parent == Course.name)
    .select(Course.name).distinct()
    .orderby(Course.published_on, order=frappe.qb.desc)
    .limit(3)
)
```

**Desktop**
- Frappe Cloud hosted instance
- Browser: Chrome
- Frappe LMS latest version

Image
Image

Sandeepambala:
The moderator fallback in `get_created_courses()` runs the base query without the instructor filter. Since the query joins `Course Instructor` with `LMS Course`, a course with multiple instructors returns duplicate rows. With `limit(3)`, a single course can occupy all slots in the "Created Courses" section.

Add `.distinct()` to ensure each course is returned only once.

Fixes #2476

### DIFF
--- a/lms/lms/api.py
+++ b/lms/lms/api.py
@@ -2101,6 +2101,7 @@ def get_created_courses():
 		.join(Course)
 		.on(CourseInstructor.parent == Course.name)
 		.select(Course.name)
+		.distinct()
 		.orderby(Course.published_on, order=frappe.qb.desc)
 		.limit(3)
 	)


raizasafeel:
Could you please resolve greptile's comments? (comment under the issue if you feel like you disagree)

raizasafeel:
Closing due to inactivity, please reopen with requested changes

PR #2564, opened 2026-07-13
fix(quiz): restore questions after lesson navigation

### LINKED ISSUE #2559: Quiz shows "0 questions" after Next → Previous
**Describe the bug**
The quiz might display "This quiz consists of 0 questions." / "This quiz has no questions available yet." in specific conditions even tho the quiz has questions.

**To Reproduce**
Open a quiz lesson, go on a quiz click Next, then Previous. The quiz re-renders with "This quiz consists of 0 questions." / "This quiz has no questions available yet." and no Start button. A full page reload fixes it.

**Expected behavior**
Next -> previous show again the "start quiz" button.

 - OS: Linux / MacOS
 - Browser: Chrome and safari
 - Version: Chrome 149.0.7827.155; Safari: 26.5

**Additional context**
Claude debugged the problem and pointed to this root cause:
## Root cause
 
`frontend/src/components/Quiz.vue` (upstream frappe/lms, present in v2.58.1 **and current main**) combined with frappe-ui's resource cache:
 
1. `Quiz.vue` creates its resource with `cache: ['quiz_with_questions', props.quizName]`. The question list is **not** part of the resource data — it lives in component-local refs (`questionsByName`, `questions`) that are populated only by the resource's `transform` / `onSuccess` callbacks.
2. In frappe-ui's `createResource` (`src/resources/resources.js`), a cache **hit** returns the *previously created resource instance* and just calls `reload()` on it. The new options — including the new component's `transform`/`onSuccess` closures — are ignored.
3. So on remount (Previous), the reload runs the **old, unmounted** component's closures, which write to the old component's dead refs. The new component's `questionsByName` stays `{}`.
4. The new component's `watch(() => quiz.data)` fires after reload and calls its own `populateQuestions()`, which filters:
   `rawQuestions.filter(row => row?.question && questionsByName.value[row.question])`
   With an empty map, **every row is dropped** → `questions = []` → "0 questions", no Start button.
First mount works because it's a cache miss: the resource is created with that component's live closures. Reload works because the cache map is in-memory per page session.
 
**Confirming experiment:** Quiz 2 visited for the first time via SPA navigation → renders 8 questions fine; Next → Previous on it → 0 questions. Reproduced identically on two quizzes.

## When the regression was introduced
 
Two upstream frappe/lms commits, both recent:
 
1. **`d0d3c855f7` — "perf(quiz): batch question fetch into one round trip" (2026-05-26), first shipped in v2.55.0 (2026-06-08).** Replaced `frappe.client.get` + per-question `get_question_details` calls with one `get_quiz_with_questions` call, and introduced the `questionsByName` side-channel ref populated inside the cached resource's `transform`. This created the dead-closure-on-remount defect. Symptom at this stage was subtler: after away-and-back navigation the count and Start button were still correct (`populateQuestions` read only `quiz.data.questions`), but question bodies wouldn't render after Start (empty map → `questionDetails.data` null).
2. **`9ae9761243` — "fix(quiz): tolerate dangling question refs in inline lesson quiz" (2026-06-19), first shipped in v2.57.0 (2026-06-24).** Added the filter `rawQuestions.filter(row => row?.question && questionsByName.value[row.question])` to `populateQuestions`. With the empty map on remount, this drops *every* row — producing the visible "This quiz consists of 0 questions." / "This quiz has no questions available yet." regression.
Version check (marker strings in Quiz.vue per tag): ≤ v2.54.2 (2026-05-20) old code, remount-safe · v2.55.0–v2.56.1 latent defect only · **v2.57.0 first release with the "0 questions" bug** · v2.58.0/v2.58.1 (deployed) affected · upstream `main` still affected as of 2026-07-10.

## Suggested patch
```
Subject: [PATCH] fix(quiz): quiz renders "0 questions" after navigating away and back
 
The quiz resource used cache: ['quiz_with_questions', quizName]. On a
cache hit, frappe-ui's createResource returns the previously created
resource instance and reloads it — the new component's transform /
onSuccess closures are never attached. The reload therefore populates
the *unmounted* component's questionsByName ref, the remounted
component's map stays {}, and populateQuestions() filters out every
question row → "This quiz consists of 0 questions." / no Start button.
 
Removing the cache option costs no extra request (a cache hit already
triggered reload()), and guarantees each mount owns its closures.
 
Applies to frappe/lms v2.58.1 (also applies cleanly to current main).
---
--- a/frontend/src/components/Quiz.vue
+++ b/frontend/src/components/Quiz.vue
@@ -556,10 +556,6 @@ const quiz = createResource({
 	makeParams() {
 		return { quiz: props.quizName }
 	},
-	// Cache key intentionally distinct from the old frappe.client.get cache
-	// — the response shape changed (now { quiz, questions_by_name }), and a
-	// stale entry would break the transform.
-	cache: ['quiz_with_questions', props.quizName],
 	auto: true,
 	transform(data) {
 		const quizDoc = data?.quiz || {}
 ```

Dkm0315:
## What changed

- keep the quiz resource instance-local so each lesson mount updates its own question map and timer state
- add a Vitest regression that faithfully reproduces frappe-ui's cached-resource callback behavior across unmount/remount
- preserve dangling-question filtering and assert that starting the quiz does not introduce per-question requests

## Root cause

`createResource` returns the original cached resource for the same quiz key and reloads it. That resource's `transform` and `onSuccess` callbacks still close over the first `Quiz.vue` instance. After lesson navigation remounts the component, the new instance's `questionsByName` remains empty, so `populateQuestions()` treats every linked question as unresolved and renders a zero-question quiz.

The cached path already performs a reload on every mount. Keeping this resource instance-local therefore fixes the stale callbacks without increasing network traffic: there is still one quiz request per mount.

### Red/green regression

- with the original cache line restored, `Quiz.test.ts` fails on the second mount: expected `1` question, received `0`
- with this fix, the same test passes and verifies exactly two batched quiz requests across two mounts, with no additional request after Start

### Automated and Bench checks

- `yarn test src/tests/Quiz.test.ts` — 1 test passed
- `yarn test --run` — 29 files, 197 tests passed
- `yarn prettier --check src/tests/Quiz.test.ts` — passed
- `git diff --check` — passed
- `bench build --app lms` — passed
- `bench --site lms-2559.localhost migrate` — passed

### Live browser acceptance

Tested on a fresh Frappe v16 bench with an isolated two-lesson fixture (quiz lesson followed by a navigation target), without reloading between navigation steps.

| Step | Unchanged code | This fix |
| --- | --- | --- |
| Open quiz lesson | 1 question; Start visible | 1 question; Start visible |
| Click Next | Navigation Target opens | Navigation Target opens |
| Click Previous | 0 questions; empty state; no Start | 1 question; Start visible |

The bounded post-fix server trace recorded exactly two `lms.lms.utils.get_quiz_with_questions` requests (initial mount and remount), both HTTP 200.

Fixes #2559

@pateljannat, this is tested and ready for review. I kept the production change scoped to this resource and included both a mutation-tested regression and live Bench evidence.


Dkm0315:
### Live Browser evidence

Same isolated fixture and the same SPA tab; no full-page reload between navigation steps.

**Before — unchanged develop:** Next → Previous remounts the quiz as **0 questions** and Start disappears.

Before fix: returning to the quiz shows 0 questions

**After — fixed branch:** the same sequence remounts it as **1 question** with Start available.

After fix: returning to the quiz restores 1 question and Start

Network trace: exactly **2** successful get_quiz_with_questions POSTs, one per mount.

PR #2594, opened 2026-07-24
feat: allow chart colors to be themed via CSS custom properties (--chart-1..--chart-5)

giovanespinach:
### Problem

The frontend can be restyled almost entirely through CSS custom properties, but chart series colors can't:

- `AxisChart`/`DonutChart` call sites that pass no `colors` (Statistics page, admin batch dashboard) fall back to ECharts' built-in palette, which no CSS can reach.
- Pages that do pass colors hardcode a per-page list (course dashboard pie, program progress summary donut).

So a theme with no blue in it still gets blue charts.

### Solution

Opt-in theme tokens, with the current behavior kept intact as fallback:

- New util `frontend/src/utils/chartColors.ts`: `getChartColors(count)` reads `--chart-1` .. `--chart-5` from the root element's computed style and returns the defined values in order — or `undefined` when none are defined.
- Call sites only pass `colors` when tokens exist:
  - `Statistics.vue` — 3× AxisChart + 1× DonutChart get `colors` via conditional spread.
  - `AdminBatchDashboard.vue` — AxisChart, same conditional spread.
  - `ProgramProgressSummary.vue` — DonutChart prefers tokens; the existing `var(--red-400)`/`--amber-400`/`--pink-400`/`--blue-400`/`--green-400` list stays as the fallback.
  - `CourseDashboard.vue` — `progressColors` prefers tokens with the existing red/amber/blue/green list as fallback. The legend dots next to the pie now read from the same `progressColors` by index, so legend and pie stay in sync when a theme overrides the tokens (identical output by default: the backend returns the distribution in a fixed order that matches the previous name-based mapping in `get_progress_distribution`).

When no `--chart-N` token is defined, nothing changes — the spread contributes nothing and the fallback lists are the ones already in the code.

Notes:
- Values are resolved through `getComputedStyle`, so any valid CSS color works, and ECharts receives concrete colors (it can derive emphasis/dimmed variants from them).
- Tokens are read once at component setup.

Context: we (LetItGrowth) run the LMS in production re-themed entirely via CSS custom properties; the charts are currently the only surface that theming approach cannot recolor, which is what motivated this change.

`yarn build` passes; touched files verified against Prettier (no formatting drift on the changed lines).

### DIFF
--- a/frontend/src/pages/Batches/components/AdminBatchDashboard.vue
+++ b/frontend/src/pages/Batches/components/AdminBatchDashboard.vue
@@ -169,6 +169,7 @@
 								type: 'bar',
 							},
 						],
+						...(chartColors ? { colors: chartColors } : {}),
 					}"
 				/>
 
@@ -209,6 +210,7 @@ import {
 import { computed, inject, ref, watch } from 'vue'
 import type dayjsType from 'dayjs'
 import { formatAmount } from '@/utils'
+import { getChartColors } from '@/utils/chartColors'
 import BatchFeedback from '@/pages/Batches/components/BatchFeedback.vue'
 import BatchStudentProgress from '@/pages/Batches/components/BatchStudentProgress.vue'
 import NumberChartGraph from '@/components/NumberChartGraph.vue'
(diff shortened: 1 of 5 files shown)


raizasafeel:
We are still debating if this is necessary from code. Closing it for now

PR #2642, opened 2026-08-01
fix: secure send_confirmation_email against untrusted payload

AhmedAbokhatwa:
**1- Fixed the type hint (Document → Document | str)** 

- since the form's JS always sends frm.doc as a JSON string  not an object

**2- Fixed a security issue:** 

`doc = frappe._dict(json.loads(doc))`
- **the old code**  took the whole JSON sent by the client and used it directly as the document, no DB lookup at all. So "trusted the entire client-sent payload as if it were the real document".

-  Now we only take the name from the payload and use it to fetch the real document from the database via frappe.get_doc  so the data actually used comes from the database, not from the client


raizasafeel:
Can you add PR description? @AhmedAbokhatwa

janhoy:
Status on this?

AhmedAbokhatwa:
Hi @raizasafeel  @janhoy  apologies for the delay, the description added  now

raizasafeel:
### Code review

Found 1 issue:

1. `name` is taken from the client payload and passed to `frappe.get_doc()` with no type check (CLAUDE.md says "Never expose `frappe.whitelist()` methods without explicit input type validation", and `specs/security.md` §1 requires "an explicit `isinstance` check (or equivalent) for every parameter at the top of the function").

`frappe.parse_json(doc).get("name")` returns whatever JSON value the caller put under that key, including an object or an array. In frappe, `Document.__init__` assigns the second positional arg to `self.name`, and `load_from_db` only takes the name-lookup path when it is `str | int` — otherwise it falls through to `frappe.db.get_value(doctype=..., filters=self.name, fieldname="*")`, where a dict or list is used as a **query filter**. So a payload such as `{"name": {"confirmation_email_sent": 0}}` loads whichever enrollment the filter happens to match. Since the `is_admin` branch below passes for any Moderator or Batch Evaluator, that lets a crafted payload send mail for, and set `confirmation_email_sent` on, an enrollment the caller never named.

(A missing `name` is handled safely — `filters=None` makes `get_values` return `None` and `load_from_db` raises `DoesNotExistError`. The gap is the non-string case.)

A guard before the lookup is enough, e.g. `if not isinstance(name, str) or not name: frappe.throw(_("Invalid enrollment"))`.

https://github.com/frappe/lms/blob/921ebeab9375b38927bccbf9a847f5521ce0c589/lms/lms/doctype/lms_batch_enrollment/lms_batch_enrollment.py#L128-L131

lms/lms/doctype/lms_batch_enrollment/lms_batch_enrollment.py
```diff
@@ -124,9 +124,10 @@ def add_member_to_live_class(self):
 
 
 @frappe.whitelist()
-def send_confirmation_email(doc: Document):
+def send_confirmation_email(doc: Document | str):
 	if isinstance(doc, str):
-		doc = frappe._dict(json.loads(doc))
+		name = frappe.parse_json(doc).get("name")
```
raizasafeel, line 129: `if not isinstance(name, str) or not name: frappe.throw(_("Invalid enrollment"))`

Could you just add this guard here? Cause otherwise it raises a `DoesNotExistError` on missing name

PR #2643, opened 2026-08-02
fix: align Save button with Not Saved badge in Edit Profile modal

FawazAlhafiz:
## Problem

In the **Edit Profile** modal, the `Save` button sits noticeably higher than the `Not Saved` badge next to it, so the header row looks misaligned.

save-button-edit-profile

## Cause

The button was wrapped in a `div` with `pb-5`. That padding made the wrapper taller than the badge beside it, and the row's `items-center` then centered the taller box — leaving the button offset upward by roughly half the padding.

The same wrapper also carried `float-end`, which has no effect on a flex child.

## Fix

Remove the wrapper so the `Button` is a direct child of the flex row. Both it and the badge then align on the same axis via the existing `items-center`.

No behavior change — purely visual.

### DIFF
--- a/frontend/src/components/Modals/EditProfile.vue
+++ b/frontend/src/components/Modals/EditProfile.vue
@@ -9,11 +9,9 @@
 					<Badge v-if="isDirty" theme="orange">
 						{{ __('Not Saved') }}
 					</Badge>
-					<div class="pb-5 float-end">
-						<Button variant="solid" @click="saveProfile()">
-							{{ __('Save') }}
-						</Button>
-					</div>
+					<Button variant="solid" @click="saveProfile()">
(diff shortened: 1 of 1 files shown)


FawazAlhafiz:
### after

image_2026-08-02_210606818

FawazAlhafiz:
Superseded by #2645.

While looking at the dialog's close (`X`) button, I found it was misaligned for a related but separate reason — `mb-5` on the header row. Since flexbox aligns margin boxes, that margin lifted the whole title block above the close button.

The two offsets push in the same direction, so removing only the `pb-5` wrapper (this PR) still leaves the `X` misaligned. #2645 removes both and fixes the header row as a whole.

PR #2647, opened 2026-08-04
fix(ui): make forms and settings list consistent

raizasafeel:
## What

Every settings panel drew its own list, and every form drew its own field labels. This replaces both with one shared implementation.

- **`SettingsList`** — new component for settings list using frappe ui's `List` (five column types, search, page size 13, Load More). All lists except raven moved to it (will be redoing raven later).
- **Labels** — one import surface for frappe-ui's labeling primitives; every control now takes the same labeling API, and the hand-drawn `FormLabel`s are gone.
- **bump frappe-ui `1.0.0-beta.29`** — plus the four call-site fixes the bump needs (`clearSelection` → `clear`, MultiLink slot props, nullable create value, narrowed combobox option types).
- **Consistency pass** — one row height and one avatar gap across the lists, columns left-aligned, one icon stroke weight, the paragraph scale in the chrome, the three Dialogs off their deprecated API.
- **RTL gate armed** — every `tailwind-rtl` semgrep rule is now an error, and the remaining physical Tailwind classes were replaced.

Also fixed along the way: the sidebar's apps and import menus were unreachable, and the command palette's shortcut hints were invisible.

## Screenshots

| Area | Before | After |
|---|---|---|
| Settings list (Categories) | image| image |
| Settings list (Zoom) | image | image|
| Users | image | image |
| Form labels | image | image|
| Sidebar apps / import menus | image | image |


raizasafeel:
@greptile-apps

PR #2649, opened 2026-08-04
fix: prevent duplicate pending LMS payments

AhmedAbokhatwa:
Problem: Multiple LMS Payment records were created for abandoned payment attempts.
Solution:Reuse existing pending LMS Payment records for the same member and document instead of creating duplicates.
Test:Added regression test test_duplicate_pending_payment.


raizasafeel:
Could you resolve the semgrep and greptile?

AhmedAbokhatwa:
@raizasafeel ٍSure I will Resolve it  Thanks for pointing them out!

AhmedAbokhatwa:
@raizasafeel  could you please review this PR?

raizasafeel:
### Code review

Found 5 issues:

1. `payment_doc.save()` in the constraint path drops `ignore_permissions=True`, which every previous save in this function had — this breaks checkout for every non-admin buyer (bug due to `lms/lms/doctype/lms_payment/lms_payment.json` granting `create`/`write` only to `System Manager` and `Moderator`, with no `if_owner` and no `has_permission`/`permission_query_conditions` entry for `LMS Payment` in `lms/hooks.py`). `record_payment` runs as the purchasing member via the whitelisted `get_payment_link`, so `Document.insert()`'s `check_permission("create")` raises `PermissionError`. The redis-lock branch above still passes `ignore_permissions=True`. Tests run as Administrator, so CI does not catch this.

https://github.com/frappe/lms/blob/1926714c96f3a15466e03fbc92461c5e439e7507/lms/lms/payments.py#L229-L232

2. `frappe.db.rollback(savepoint=...)` uses the wrong keyword — the parameter is `save_point` (bug due to `def rollback(self, *, save_point=None, chain=False)` in `frappe/database/database.py`). This raises `TypeError` inside the `except` block, replacing the unique-key violation it was meant to handle with an unhandled error. This is the path that runs whenever two checkouts actually race, which is the case the PR exists to handle.

https://github.com/frappe/lms/blob/1926714c96f3a15466e03fbc92461c5e439e7507/lms/lms/payments.py#L233-L236

3. `frappe.cache().redis.set(...)` raises `AttributeError` (bug due to `frappe.cache` being a `RedisWrapper(redis.Redis)` whose `__call__` returns `self`; `.redis` is an attribute of the separate `ClientCache` class in `frappe/utils/redis_wrapper.py`, not of `RedisWrapper`). This branch runs whenever `has_unique_pending_payment()` is false — and that is permanent on a site with pre-existing duplicates, because the fallback adds an index named `unique_pending_payment_lookup` while `has_unique_pending_payment()` looks for `unique_pending_payment`. So the sites the fallback exists for get a 500 on every checkout. `frappe.cache().set_value(...)` is the idiomatic call.

https://github.com/frappe/lms/blob/1926714c96f3a15466e03fbc92461c5e439e7507/lms/lms/payments.py#L215-L218

4. Gating `payment_doc.update(payment_details)` on `is_new()` means a reused pending row keeps the first attempt's data. On a retry with a different coupon, amount or address, the stored `amount`/`coupon`/`discount_amount`/`address`/`gstin` stay stale while `get_payment_link` recomputes `total_amount` fresh for the gateway, so the row no longer matches what is charged — and `update_coupon_redemption` and `complete_enrollment` read that row. The redis-lock branch updates unconditionally; the doc re-fetched in the unique-violation handler is likewise never updated.

https://github.com/frappe/lms/blob/1926714c96f3a15466e03fbc92461c5e439e7507/lms/lms/payments.py#L226-L229

5. `test_duplicate_pending_payment` calls `_buy_course()` twice sequentially, so it never exercises the race (`specs/specs-ankush/concurrency.md` rule 5 says "Add a test that runs the endpoint twice in parallel (real threads or processes) and asserts only one row exists"). The second sequential call finds the existing row and takes the update path, so the `except` branch in issue 2 is never reached. `threading` is already imported in this file, and `test_coupon_redemption.py` has the pattern.

https://github.com/frappe/lms/blob/1926714c96f3a15466e03fbc92461c5e439e7507/lms/tests/api/test_payments.py#L147-L152

PR #2662, opened 2026-08-12
feat(mobile): forms become pages, lists become cards, filters move into a sheet, and settings leaves the phone

raizasafeel:
## Summary

**Forms** 
- Turn modals into pages. 19 create/edit forms have addresses now, so you can link straight into "new batch" and the back button means something. `useFormRoute` is the contract, `FormShell` draws it — one declaration renders a dialog on desktop and a `fixed inset-0` page below `sm`.

**Lists and filters**
- One list surface at every width 
- Filters collapse into a sheet behind one Filters button with a count badge. The slot renders in exactly one place, never both with one hidden by CSS — two renders would mount two copies of every control bound to the same page ref.
- Boolean filters go from chips back to checkboxes; `FilterChip.vue` is deleted. The chip existed because a checkbox row doesn't survive a 390px header — now that filters live in a sheet, that reason is gone, and frappe-ui's checkbox is an `<input>` plus `<label for>`, so the label is part of the hit area natively.
- The bulk-selection banner is teleported out of the list: frappe-ui pins it `absolute bottom-6` against a root as tall as every row, landing it hundreds of pixels below the screen. On a phone the footer gives up the bottom edge while a selection is open rather than sitting underneath, where its controls would still take focus out of sight.

**Navigation and You**
- The bar's fifth slot was More, a sheet of everything that didn't fit. It's now You, drawn as the user's own avatar rather than a glyph, and the primaries change from Home/Courses/Certifications/Profile to Home/Courses/Batches/Programs. Everything the bar doesn't hold is grouped on the You page.
- The link list moves to a module store, so You (a route of its own, reachable by cold deep link) and the bar read the same source. Runs are versioned and memoised per viewer, so a learner is asked for `get_programs` once per boot instead of once per watcher firing.

**Settings**
- Settings comes off the phone entirely. `/settings` and `/settings/:item` fall through to NotFound; `/settings/users/:memberID` is the only thing under the prefix with an address. An LMS is not configured with a thumb. Desktop keeps the whole tree.
- The 600-line settings tree lifts out of `Settings.vue` into `settingsStructure.js`, so the dialog and anything else read one source. `isSettingsMounted` tells callers whether there's a dialog listening at all — otherwise a moderator asking for Settings from a routed form on a phone had their form closed for a dialog that never appeared.

## Screenshots

1. Lists
image

2. Filters
image

3. Profile
image

4. Modals
image


raizasafeel:
@greptile-apps

PR #2664, opened 2026-08-12
fix(security): sanitize what renders and what saves, allowlist bound URL schemes, and pair every permission gate

raizasafeel:
### Problem
- `v-html` dumps a string into the page as real HTML.
- If that string came from a user, they can put `` in it.
- Their code then runs on our domain, as the logged-in user.
- One bad course description hits everyone who opens that page.

### Fix
- Drop-in replacement. Same job, but cleans the HTML first (DOMPurify).
- Written as `v-safe-html:rich="text"` instead of `v-html="text"`.
- The word after the colon is the level.

#### The three levels

- **`rich`** — real authored content: lesson bodies, course and batch  descriptions, quiz questions, assignment questions, job descriptions.
  - Blocklist. Keeps everything except form tags (`form`, `input`, `button`, …).
  - Tables, images, code blocks, headings and styles all survive intact.
- **`basic`** — notification subjects only.
  - Allowlist, short. Drops image alt text, `colspan`, classes and styles.
- **`bio`** — user bios.
  - Strictest allowlist. Prose and links only; no headings, tables or code.
- **A missing or misspelt level falls back to `bio`.**
  - Too strict is a visible bug someone reports. Too loose is an XSS nobody sees.

## The CI rule

- `no-raw-v-html` fails the build on any `v-html=` in a `.vue` file.
- `no-bound-innerhtml-prop` fails on `:innerHTML=` too — same thing under another
  name, and banning only one leaves a documented bypass.
- `.vue` only, deliberately: `v-safe-html` is a Vue directive, so in a `.js` file
  the rule would demand a fix that cannot be written. Those files call
  `sanitizeAt(level, value)` directly instead.
- Currently 0 findings across 234 files, verified against a positive control.

## Everything else in the PR

- **`safeUrl`** — a `javascript:` URL in the lesson media-macro iframe `src` ran
  in our origin, for Guest too. 42 bound `src`/`href` across 22 components are
  now scheme-checked.
  - A rejected URL returns `undefined`, not `''`, so Vue removes the attribute.
    `src=""` would reload the whole SPA inside the iframe.
- **Embeds no longer vanish** — a stripped `<iframe>` used to leave nothing
  behind. It now degrades to a link to its source.
- **Saving no longer destroys content** — the short-field allowlist has no `alt`,
  `scope`, `colspan`, `rowspan` or `caption`, and 9 of 10 `sanitizeHTML` calls
  saved their result. Assignment tables and image alt text were being deleted on
  save, permanently. Saving is security-only now.
- **Permissions** — 4 doctypes checked their rule on single-doc reads but not on
  list reads. The missing query conditions are added.
- **Two snapshot tests** freeze the low-privilege DocPerms and the 26
  `allow_guest` endpoints, so a new one has to arrive in a reviewed file.
- **New endpoints** — doctype permissions in `get_user_info`,
  `get_doc_permissions_many`, and a `usePermissions` composable that answers
  "no" until the server has actually replied.
- **Enrolment race** — `exists()` then `insert()` with no lock meant two
  concurrent requests both inserted. Fixed with unique indexes.
- **3 semgrep rules** for the audit's repeat offenders: GET-reachable mutators,
  unjustified `ignore_permissions`, read-then-insert.
- **Badge "Assign For" never saved** — frappe-ui's Combobox emits a plain value,
  the handler read `opt.value` off it, got `undefined`, and cleared the field.
  Every badge saved without the field that decides what it is awarded for.
- **White text on pale backgrounds** — three spots at about 1.1:1, effectively
  invisible.

### DIFF
--- a/.github/semgrep/fixtures/security.py
+++ b/.github/semgrep/fixtures/security.py
@@ -1,0 +1,48 @@
+"""Known positives for security.yml. Every function here MUST be flagged,
+apart from the suppressed one, which proves the escape hatch works.
+
+The filename has to match the rule file's stem for `semgrep --test`.
+Not imported by the app; excluded from coverage."""
+
+import shutil
+
+import frappe
+
+
+# ruleid: lms-mutating-whitelist-needs-post
+@frappe.whitelist()
+def mutating_on_get(name: str):
+	frappe.delete_doc("LMS Course", name)
+
+
+# ruleid: lms-mutating-whitelist-needs-post
+@frappe.whitelist()
+def rmtree_on_get(path: str):
+	shutil.rmtree(path)
+
+
+@frappe.whitelist()
+def bypass_without_reason(course: str):
+	# ruleid: lms-unjustified-ignore-permissions
+	return frappe.get_all("LMS Course", filters={"name": course}, ignore_permissions=True)
+
+
+@frappe.whitelist()
+def bypass_via_flags(course: str):
+	doc = frappe.get_doc("LMS Course", course)
+	# ruleid: lms-unjustified-ignore-permissions
+	doc.flags.ignore_permissions = True
+	return doc
+
+
+def bypass_with_reason(course: str):
+	# ok: lms-unjustified-ignore-permissions
+	# nosemgrep: lms-unjustified-ignore-permissions - background job, no user session
+	return frappe.get_all("LMS Course", filters={"name": course}, ignore_permissions=True)
+
+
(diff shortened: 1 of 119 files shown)


raizasafeel:
@greptile-apps

raizasafeel:
@greptile-apps

raizasafeel:
@greptile-apps

raizasafeel:
@greptile-apps

PR #2674, opened 2026-08-19
feat: enforce lesson completion

### LINKED ISSUE #2178: Enforce Lesson Completion

raizasafeel:
## Summary

- Adds a per-course `Enforce Lesson Completion` setting. When it is off, a student can only open a lesson once every lesson before it in the course outline is complete. Off by default, so existing courses are unaffected. 
- The gate is enforced server-side at every route that can reach lesson content, rather than hidden in the UI.

### Locked Lessons on Sidebar
Screenshot From 2026-08-19 11-22-23

### Locked Lessons in the Course Outline
image

### The Locked Lesson Banner
image

### A Lesson That Does Not Exist
image

### The Course Setting
image

## How

**The rule is one pure function.**

-  `utils.compute_locked_lessons(ordered_lesson_names, completed) -> set` takes the flattened outline and the set of completed lessons and returns the locked ones. 
- No DB access, no request state — it is directly unit-testable, and every enforcement point derives from it
- `permissions.py` wraps it for callers:

| function | returns |
|---|---|
| `enforces_lesson_completion(course)` | is the setting on for this course |
| `get_locked_lessons(course)` | the locked set for the current user |
| `get_lesson_gate(course)` | the locked set **and** the lesson to send the student to |

`get_lesson_gate` derives the redirect target from the same ordered list as the lock set, and only honours `LMS Enrollment.current_lesson` when that pointer is itself unlocked — a stale pointer can otherwise aim the student at a locked lesson and dead-end the course.

**Four enforcement points, each with a bypass test:**

- `get_lesson` — returns a `locked` payload instead of lesson content
- `SCORMRenderer._check_permission` — the SCORM page is its own route and never calls `get_lesson`, so it needs its own check
- `save_progress` — it writes the very rows that unlock the gate
- `can_access_quiz` — a lesson's quiz is reachable by its own id

`build_outline` also stamps per-lesson `locked` so both outline components can render the lock and stop navigating, and it withholds a locked SCORM chapter's launch file.

**Frontend.** `LockedLessonNotice.vue` replaces the lesson body with a banner explaining why, plus a counter and a `Go now` button back to the lesson the student is actually on. `ChapterRow.vue` and `StudentLessonSidebar.vue` render locked rows as non-navigating, named with `sr-only` text.

## Issues
- Closes #2178

### DIFF
--- a/frontend/components.d.ts
+++ b/frontend/components.d.ts
@@ -91,6 +91,7 @@ declare module 'vue' {
     ListPage: typeof import('./src/components/Layouts/ListPage.vue')['default']
     LiveClassAttendance: typeof import('./src/components/Modals/LiveClassAttendance.vue')['default']
     LMSLogo: typeof import('./src/components/Icons/LMSLogo.vue')['default']
+    LockedLessonNotice: typeof import('./src/components/LockedLessonNotice.vue')['default']
     LucideChevronLeft: typeof import('~icons/lucide/chevron-left')['default']
     LucideChevronRight: typeof import('~icons/lucide/chevron-right')['default']
     LucideStar: typeof import('~icons/lucide/star')['default']
--- a/frontend/src/components/ChapterRow.vue
+++ b/frontend/src/components/ChapterRow.vue
@@ -13,7 +13,10 @@
 			/>
 			<div
 				class="ms-2 min-w-0 flex-1 text-start"
-				:class="inlineSelect ? '' : 'flex items-baseline justify-between gap-3'"
+				:class="[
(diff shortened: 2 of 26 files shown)


raizasafeel:
```
The PR is not safe to merge until legacy public SCORM files are covered by the sequential-completion authorization boundary.

Production nginx still serves legacy public SCORM assets without invoking the newly added renderer lock check, so students can directly access locked package content.

```

not created by the PR so will be pushing another fix separately

PR #2697, opened 2026-08-31
 Resolve attributeError in count_matching

LeoDanielA01:
When opening the Batches page, the footer was supposed to show 
how many batches match the current filters (like "8 of 24"). 
Instead it was throwing an error and nothing would load.

The issue was in how we were counting the batches behind the 
scenes. We were using an older way of doing it that no longer 
works with the current version of Frappe. 

Updated the count_matching function in utils.py to use a 
different approach that works correctly,  it now builds the 
filtered query first and runs a count on top of it, so the 
footer shows the right number every time without crashing.


raizasafeel:
Merged, thanks for the contribution!

PR #2698, opened 2026-08-31
Improve text contrast in programming exercise submission

LeoDanielA01:
Two small visual fixes on the Programming Exercise Submission page.

**1. Run button** : removed a hardcoded gray text color that was 
overriding the button's default white label.

**2. Compiler error message** : the error text was too light to read 
comfortably. Increased the red shade so it stands out more clearly 
when something goes wrong.

### Before:
image

image

### After:
image

image


raizasafeel:
Merged! Thanks for the contribution!

PR #2707, opened 2026-09-07
fix(lms): deduplicate course filters for created and enrolled courses

### LINKED ISSUE #2476: [Bug] get_created_courses() moderator fallback missing .distinct() causes duplicate course cards on Home page

Meghwin-Dave:
Fixes #2476. Ensures created and enrolled course lists use unique course names when building the filter query to prevent duplicate course cards from rendering for instructors/moderators with multiple batch assignments.

### DIFF
--- a/lms/lms/utils.py
+++ b/lms/lms/utils.py
@@ -1025,14 +1025,14 @@ def update_course_filters(filters: dict) -> tuple:
 
 	if filters.get("enrolled"):
 		enrolled_courses = frappe.get_all("LMS Enrollment", {"member": frappe.session.user}, pluck="course")
-		filters.update({"name": ["in", enrolled_courses]})
+		filters.update({"name": ["in", list(set(enrolled_courses))]})
 		del filters["enrolled"]
 
 	if filters.get("created"):
 		created_courses = frappe.get_all(
 			"Course Instructor", {"instructor": frappe.session.user}, pluck="parent"
 		)
-		filters.update({"name": ["in", created_courses]})
+		filters.update({"name": ["in", list(set(created_courses))]})
 		del filters["created"]
 
 	if filters.get("live"):
--- a/lms/tests/test_utils.py
+++ b/lms/tests/test_utils.py
@@ -518,6 +518,12 @@ class TestListEndpointPaging(BaseTestUtils):
 	def test_the_count_includes_the_featured_courses(self):
 		self.assertEqual(get_course_count(filters=self._filters()), 4)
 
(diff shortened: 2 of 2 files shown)


raizasafeel:
Closing this, the change doesn't actually fix anything and the test doesn't validate what it claims to.

- `update_course_filters`/`get_courses` do a single-table read (`filters=... IN (...)`), no join — duplicate values in an `IN` list can't produce duplicate output rows, so `list(set(...))` is a no-op here.
- The test can't fail without the change: fixtures give every course one instructor (so the "multiple instructors" case never gets built), and it calls `get_courses` with no category filter, so on a fresh CI site it may just read zero rows.
- The real duplicate-course bug lives in `get_created_courses`/`get_created_batches` (they join `Course Instructor`, so a 3-instructor course comes back 3x), already patched in #2686.
- Side note: unrelated proctoring/Quiz.vue comments here are as you raised it against main (we work against develop branch)

PR #2712, opened 2026-09-08
fix: reveal the signup result banners on released frappe

FawazAlhafiz:
### Problem

#2691 inlined the `alert_banner` macro so `/login` stops returning a 500 on a released frappe. That part works — but the banners it renders never become visible. Sign-up succeeds or fails and the user sees nothing: the message lands in the DOM, correctly, and stays invisible.

Two independent causes:

**1. Released frappe hides them by default.** `login.bundle.scss` styles `.login-error-banner` / `.login-success-banner` completely — padding, gap, background, radius, colour, svg fill — but sets `display: none`, expecting its own JS to flip it. LMS only removed the `hidden` class, which does not touch `display`.

**2. frappe's own JS re-hides them.** `login.js` binds a page-wide handler:

```js
$(".page-card-body input").on("input", function () {
    $(this).closest(".page-card-body").removeClass("invalid").find(".login-error-banner").hide();
});
```

The LMS form has its own `.page-card-body`, so it is caught by that selector. jQuery's `.hide()` writes an **inline** `display: none`, which outranks any stylesheet rule — so once the user has typed their email, no class toggle can reopen the banner.

### Fix

Reveal with `.css("display", "flex")` rather than a class toggle. This is what frappe's own `login.show_error_banner` does, and being inline it clears the stamped `display: none` and outranks the stylesheet in a single step — no specificity war, and no assumption about which frappe version is running.

The icon needed two smaller corrections:

- **`width="16" height="16"` attributes** — neither frappe version sizes the banner svg in CSS; released frappe's own login markup carries these same attributes.
- **`fill` moved from an attribute to an inline style** — frappe's `.login-*-banner svg { fill: var(--ink-red-4) }` overrides a presentation attribute, which fills in the stroked lucide glyph and renders the icon as a solid dot.

No CSS is added. Both frappe versions already style these banners (released via `login.bundle.scss`, develop via `es-alert`); the only thing neither supplies is `display`.

### Screenshots

**Error path — existing email**

signup-error-banner

**Success path — new email**

signup-success-banner

### Tests

Two regression tests added to `lms/tests/test_signup_form.py`, pinning the two decisions a future cleanup would most plausibly undo — the explicit `display`, and the inline `fill` / sized icon.

Note that true visibility can only be asserted in a browser, and this repo has no guest-flow e2e harness (the cypress specs all assume a logged-in author). These tests guard the source-level decisions; the rendering itself was verified manually against a released frappe.

### Verification

Checked end to end against frappe v16 on a live site, driving the real form: both banners report `display: flex`, `isVisible: true`, correct text, and correctly rendered outline icons.

### DIFF
--- a/lms/templates/signup-form.html
+++ b/lms/templates/signup-form.html
@@ -1,7 +1,7 @@
 {%- macro alert_banner(type, text="") -%}
 {%- set is_error = type == "error" -%}
 <div class="login-{{ type }}-banner es-alert hidden" role="alert" data-theme="{{ 'red' if is_error else 'green' }}">
-    <svg fill="none" stroke="currentColor" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round" aria-hidden="true">
+    <svg width="16" height="16" style="fill: none" stroke="currentColor" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round" aria-hidden="true">
         <use href="#icon-{{ 'circle-alert' if is_error else 'circle-check' }}"></use>
     </svg>
     <div class="es-alert__content">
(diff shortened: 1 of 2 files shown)


FawazAlhafiz:
Hi @raizasafeel , could U take a look at this?

raizasafeel:
Seems good, merged, thanks for the contribution!

PR #2714, opened 2026-09-09
fix: rename_intrest_to_course_interest

mahsem:
Interest already exist so rename it better description


raizasafeel:
This is not an issue and if you would like to implement this customization, feel free to fork it!

PR #2722, opened 2026-09-10
feat: add google api, calendar, badge assignments; redo settings ui

### LINKED ISSUE #2227: Google Settings and Google Calendar on the Frontend
## Current Problem
Google Calendar from Framework is being used in multiple places. Evaluators must have a Google Calendar to be able to take the evaluation. Google Calendar is also needed for Google Meet live classes. But right now, there is no way to add a Google Calendar from the frontend. Also, Google Calendar needs Google Settings enabled, which is also not available on the frontend.

Also, as of now, Settings is visible only to Moderators. But batches can be created by Evaluators, too. They may want to add a new Google Meet account and add their calendar, but without the Settings dialog, they cannot do this.

## Solution
1. There needs to be a Google Section in the Settings dialog. This will have Google Settings and Google Calendar.
2. The Settings dialog should also be visible to Evaluators. This could be with restricted access. They should only be able to access the conferencing section and the Google calendar. They should also only see their calendar and Google account, and not every account on the list.

raizasafeel:
### Changes

- **Settings architecture:** restructured page chrome/layout under Layouts/pages and Layouts/settings (desktop + mobile); the settings dialog is now driven from the URL hash instead of local modal state.
- **Settings tree as data:** Coupons, Transactions, Payment Gateways, Zoom, Google Meet, Badges, and Members split into separate config+list+form pages, declared as data instead of hardcoded routes.
- **Sidebar management:** sidebar items moved from hardcoded links to stored LMS Sidebar Item rows; built-in rows seed on install/migrate, and users can manage them from Settings > Sidebar.
- **System Preferences:** exposed language, timezone, and text-direction
- **Google integration:** Google API and Google Calendar panels wired into the Conferencing settings section.
- **Communication > General:** new fields page holding contact info and the notification-template pickers, moved out of System Configuration.
- **Unified Roles UI:** one shared RoleSwitches component replaces four separately-implemented role-toggle UIs — the public profile's Settings tab, the desktop and mobile Members edit forms, and the Add/Edit Member dialog — each previously its own layout and role list.
- **Badge Assignment**: exposed on frontend

### Why these changes

- One declared settings tree + shared record-page composables cuts duplication and makes new settings pages easier to add.
- A data-driven sidebar lets admins customize navigation without a code change.
- URL-hash-driven dialogs tie settings state to browser history for correct back-button behavior.
- Unifying the roles UI removes four drifted implementations of the same control.

### Bugfixes included

- Services.vue save no longer throws TimestampMismatchError (secret write now sequenced after the record save, not concurrent).
- Email Template link-field pickers no longer show "No results" (a CRM-added enabled filter was silently excluding every template).
- List pagination offset restored correctly after a failed Load More.
- Job-listing rate limiting, RTL language source, and batch timezone-default fixes.

#### Quick comparison

| Before | After |
| ---- | ---- |
| image |image|

#### Video Walkthrough

https://github.com/user-attachments/assets/73c191ce-383f-415d-be65-ee405e6d295b

### Issues
- Closes #2227


raizasafeel:
@greptile-apps

raizasafeel:
@greptile-apps

raizasafeel:
@greptile-apps

PR #2727, opened 2026-09-10
fix: change all batches tab to active and implemented dynamic title name

### LINKED ISSUE #2711: Change `all` tab in batches to `active`
- Change all batches to active batches
- This does not change student view as they only see active batches, but instead helps admin roles, right now looks very cluttered.
- This should have `upcoming + current` batches

ajzal-byte:
Resolves #2711

- replaced the "All" tab with an "Active" tab specifically for admin
- updated the updateTabFilter logic so the "Active" tab correctly displays both current and upcoming batches
- made the page title dynamic so it accurately reflects the currently selected tab


raizasafeel:
### Code review

Found 1 issue:

1. The new "Active" tab's filter has no lower bound tied to `start_date`, so it overlaps with "Upcoming": a batch that hasn't started yet (`start_date` in the future, `end_date` further out) satisfies `end_date >= today` and is shown/counted as "Active" even though it's really upcoming. Compare the sibling branches in the same function, which do bound on `start_date` (`upcoming`: `start_date >= today`, `archived`: `start_date <= today`).

https://github.com/frappe/lms/blob/fb7531b9f76c56051ed9a0b4d9ed954a20ac47fd/frontend/src/pages/Batches/Batches.vue#L264-L277

This also reintroduces, for `end_date`/`end_time`, the same class of date-only boundary bug that `get_batch_type()` / `filter_batches_based_on_start_time()` in `lms/lms/utils.py` were added to fix for `start_date`/`start_time` — those helpers key only on `start_date`, so they don't run for this new `end_date` filter shape, meaning a batch whose `end_date` is today but whose `end_time` has already passed still shows as "Active" for the rest of the day.

ajzal-byte (commented):
active filter: added the start_date <= today bound so upcoming batches no longer show up under active.

endd time logic: updated utils.py and test_utils.py to check end_time so batches ending today drop off as soon as they finish.

roles & guests: switched admin roles to a Set, allowed the "Enrolled" tab if someone has both admin and student roles, and hid tabs entirely for guests.

frontend/src/pages/Batches/Batches.vue
```diff
@@ -114,7 +114,7 @@ import { computed, inject, onMounted, ref, watch } from 'vue'
 import { useRouter } from 'vue-router'
 import { sessionStore } from '@/stores/session'
 import BatchCard from '@/pages/Batches/components/BatchCard.vue'
-import ListPage from '@/components/Layouts/pages/ListPage.vue'
+import ListPage from '@/components/Layouts/ListPage.vue'
```
raizasafeel: Why is this changed? The path given here doesnt exist

lms/lms/utils.py
```diff
@@ -3195,20 +3218,49 @@ def filter_batches_based_on_start_time(batches: list, filters: dict) -> list:
...
+
+	# Active is the intersection: started (start_date <= today) and not ended
+	# (end_date >= today). Check it first so that start_date <= today is not
+	# read as Archived.
+	if start_date_filter and end_date_filter:
+		if "<" in start_date_filter[0] and ">" in end_date_filter[0]:
+			return "active"
 
-	return batchType
+	if start_date_filter:
```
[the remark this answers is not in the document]
raizasafeel: Seems to be valid, can you check @ajzal-byte

frontend/src/pages/Batches/Batches.vue
ajzal-byte: must have slipped while i was rebasing. i've reverted it back

ajzal-byte:
### UI Screenshots

#### 1. Guest View
image

#### 2. Student View
image

#### 3. Instructor / Administrator View
image

PR #2733, opened 2026-09-14
chore: complete Russian translation

ZubikIT:
Fills all 435 empty entries in lms/locale/ru.po. The file was 77% translated (1460 of 1895); it is now complete.

Product names are kept as-is (Zoom, Google Meet, Moodle, Canvas, Kajabi, Notion, Teachable, Thinkific, TalentLMS, Frappe Learning), as are literal field names in validation messages ("name must be a string" and friends), since those refer to API fields rather than UI labels.

Also fixes three entries flagged `python-format` that actually use brace format: `{0}% complete`, `You got {0}% correct answers...` and `You will have to get {0}% correct answers...`. In the msgid the `%` happens to be followed by `c`, which msgfmt accepts as the `%c` conversion specifier, so the mistake stays invisible in English. Any translation whose next word does not start with a valid specifier letter — every Cyrillic one, for instance — then fails `msgfmt --check`. Two of those three were already broken before this change: `msgfmt --check` did not pass on ru.po as it stood. It does now.


raizasafeel:
Thanks for the contribution! Merged!

PR #2735, opened 2026-09-14
fix(assignment): reset submission view when switching assignments

### LINKED ISSUE #2734: Assignment view shows another assignment's submission (Assignment.vue keeps stale state across navigation)
## Description
On the learner assignment view, navigating from an assignment that **has** a submission to a **different** assignment shows the first assignment's submission (its uploaded file / answer / comments) on the second — even though the second assignment has no submission of its own (its View Submissions list is empty).

## Steps to reproduce
1. As a learner, submit assignment **A** (e.g. upload a Document).
2. Open A's submission view `/lms/assignment-submission/<A>/<submission>` — the file shows correctly.
3. Navigate (in-app, no full reload) to a **different** assignment **B** you have **not** submitted → `/lms/assignment-submission/<B>/new`.
4. B renders **A's** uploaded file / answer, even though B has no submission.

## Expected
B shows an empty "new" submission (it has none of its own).

## Actual
B shows A's `attachment` / `answer` / `comments` — stale state from the previously viewed assignment. It is display-only (no real submission is created for B; B's submissions list stays empty), which makes it especially confusing for learners and evaluators.

## Root cause
`frontend/src/pages/AssignmentSubmission.vue` renders `<Assignment>` **without a `:key`**, and `frontend/src/components/Assignment.vue`:
- creates `assignment` (`createResource`) and `submissionResource` (`createDocumentResource`) bound to the **initial** `props.assignmentID` / `props.submissionName`, with no `watch` to refetch when the props change;
- populates `answer` / `attachment` / `comments` refs from a submission and **never resets** them.

So when the route params change, the component is reused and keeps the previous assignment's state.

## Suggested fix
- Add a `:key` on `<Assignment>` in `AssignmentSubmission.vue` keyed by `assignmentID` + `submissionName` (forces a fresh mount per assignment/submission), and/or
- In `Assignment.vue`, `watch` the props to reset `answer`/`attachment`/`comments` and refetch `assignment`/`submissionResource` on change.

I'll open a PR with the fix + a test.

Environment: `frappe/lms` develop.

AbhishekNegi25:
edit: Partly fixes #2734

### Problem
On the learner assignment view, navigating from an assignment that **has** a submission to a **different** assignment shows the first assignment's submission (uploaded file / answer / comments) on the second — even though the second assignment has no submission of its own (its View Submissions list stays empty). It's display-only (no real submission is created), which is confusing for both learners and evaluators.

### Root cause
`pages/AssignmentSubmission.vue` renders `<Assignment>` **without a `:key`**, and `components/Assignment.vue` binds `assignment` / `submissionResource` to the initial props (no `watch` to refetch on prop change) and never resets `answer` / `attachment` / `comments`. So when the route params change, the component is reused and keeps the previously viewed assignment's state.

### Fix
Add a `:key` on `<Assignment>` (`assignmentID` + `submissionName`) so it remounts fresh whenever the assignment or submission changes — resetting all local state and refetching the correct submission. One line, no other behavior change.

### Testing
- Reproduced on a learner account (submitted assignment A, not B): opening A, then navigating to B, showed A's uploaded file on B. With the fix, B shows an empty "new" submission.
- The existing `assignmentStudentView` tests exercise the iframe-URL/Student-View path and are unaffected. Happy to add a component-remount regression test in whatever pattern you prefer.

cc @pateljannat


raizasafeel:
Please attach screenshots/screenrecordings when opening PRs of this kind where you can see the difference

raizasafeel:
This is safe to merge and fixes the main stale answer/attachment issue, but it does not fully close #2734.

`createDocumentResource` still caches the `"new"` submission resource, and after submit that cached entry can hold the previous submission’s status/comments. Opening another blank submission can therefore still show stale metadata.

I’d merge this as a partial fix, but I'm updating the description and keeping #2734 open until the `"new"` resource case is handled. Let me know if you would like to contribute the rest half!

Thanks for the contribution! Please attach a ss next time!

PR #2737, opened 2026-09-14
fix(i18n): load translations before mounting frontend

### LINKED ISSUE #2732: Frontend can mount before translations are loaded
## Summary

The Vue frontend can mount before its translation dictionary has loaded. On a hard refresh, users whose language is not English can therefore see source-language labels during the initial render. A few public-page labels are also passed to components as raw English configuration values and never reach the translation helper.

## Reproduction

Using LMS `v2.61.0` at commit `d3bfe97d178eb076310dffd7407106bcdec15d67`:

1. Set the user language to Spanish.
2. Hard-refresh `/lms/courses`, `/lms/batches`, or `/lms/statistics`.
3. Observe English labels such as empty-state entity names or chart titles despite translations being available.

## Cause

`translationPlugin` starts an asynchronous resource request while `main.js` mounts the application immediately. The first render is not synchronized with that request. In addition, some visible component configuration values are raw strings instead of calls to `__()`.

## Expected behavior

- Load the translation dictionary before mounting the Vue application.
- If the request fails, continue booting with source messages rather than blocking the application.
- Route visible labels through `__()` while preserving internal identifiers as untranslated values.

## Proposed fix

Expose a deduplicated translation-loading promise, await it during bootstrap, and translate the affected public-page, mobile-navigation, user-menu, and statistics labels. Cover successful loading, an existing dictionary, and request failure with frontend tests.

## Downstream context

- Tracking issue: https://github.com/germanegomez/clinic-platform/issues/159
- Temporary fork branch: https://github.com/germanegomez/lms/tree/clinic-platform/i18n-runtime-fixes-v2.61.0

germanegomez:
## Problem

The Vue frontend currently mounts while its translation request is still in flight. On a hard refresh, the initial render can therefore use source-language strings even when a translation dictionary is available. Some visible labels are also passed to components as raw configuration values and bypass `__()` entirely.

Fixes #2732.

## Solution

- Load and await translations before creating and mounting the Vue application.
- Reuse an existing dictionary and deduplicate concurrent loading requests.
- Continue booting with source messages if translation loading fails or stalls for 10 seconds.
- Route public empty-state names, statistics labels, and user-menu actions through `__()`.
- Keep chart series names, data keys, and backend resource parameters untranslated.

The current `develop` branch already translates the active mobile navigation implementation, so this change does not restore or modify the removed legacy layout.

## Testing

Executed with Node 24 and Yarn 1 in the same app layout used by the frontend workflows:

- `yarn vitest run src/tests/translation.test.ts src/tests/coursesFilters.test.ts` — 12 tests passed.
- `yarn test --run --maxWorkers=4` — 1,603 tests passed in 150 files.
- `yarn build` — production build passed.

The branch is based on the current `develop` tip and contains one commit affecting seven focused frontend files, with no dependency or generated-asset changes.


frontend/src/main.js
```diff
@@ -5,36 +5,43 @@ import App from './App.vue'
...
-watch(userResource, () => {
-	if (userResource.data) {
-		app.use(telemetryPlugin, { app_name: 'lms' })
-	}
-})
-
-app.config.globalProperties.$user = userResource
-app.config.globalProperties.$dialog = createDialog
+const bootstrap = async () => {
+	await loadTranslations()
```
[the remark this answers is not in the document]
germanegomez: Resolved in `22d0b7f4` by bounding translation loading to 10 seconds. If the request rejects or remains pending past that limit, the loader installs an empty dictionary and bootstrap continues with source messages.

I added a fake-timer regression test for the stalled-request path. Final validation: 12 focused tests passed, the complete suite passed 1,603 tests in 150 files with `--maxWorkers=4`, and the production build passed.

frontend/src/components/Sidebar/UserDropdown.vue
```diff
@@ -35,7 +35,7 @@
 							>
 								{{ branding.data?.app_name }}
 							</span>
-							<span v-else> Learning </span>
+							<span v-else>{{ __('Learning') }}</span>
```
raizasafeel: This is the name of the app (branding) so no need for translation. Can you revert just this part?

raizasafeel:
Also could you either resolve conflicts or rebase on top of develop and also lint it as per [contribution guidelines](https://github.com/frappe/erpnext/wiki/Contribution-Guidelines), that should resolve the failing test?

frontend/src/components/Sidebar/UserDropdown.vue
germanegomez: Agreed. Reverted just this change in 466e09b6, so the fallback remains the untranslated `Learning` brand name.

germanegomez:
Rebased onto the latest `develop`, resolved the conflicts, and ran the project lint checks successfully. The updated GitHub Actions workflows are awaiting maintainer approval.

frontend/src/translation.js
```diff
@@ -28,13 +30,45 @@ function translate(message) {
...
+export function loadTranslations() {
+	if (window.translatedMessages) {
+		return Promise.resolve(window.translatedMessages)
+	}
+	if (translationsPromise) {
+		return translationsPromise
+	}
+
+	translationsPromise = withTimeout(
+		call('lms.lms.api.get_translations'),
```
raizasafeel: This removes the existing IndexedDB-backed translation cache and replaces it with a network request on every load.

Could we read from the cache first and fall back to the network if needed? Another option would be to inline the translations into `index.html`, since `boot.lang` is already available there.

frontend/src/translation.js
```diff
@@ -28,13 +30,45 @@ function translate(message) {
...
+		.then((data) => {
+			window.translatedMessages = data || {}
+			return window.translatedMessages
+		})
+		.catch((error) => {
+			console.warn(
+				'Unable to load translations; using source messages.',
+				error
+			)
+			window.translatedMessages = {}
```
raizasafeel: On timeout this sets translations to `{}`, which is truthy. That means the check at line 34 will treat translations as loaded, even if the request finishes shortly after.

frontend/src/translation.js
```diff
@@ -1,9 +1,11 @@
-import { createResource } from 'frappe-ui'
+import { call } from 'frappe-ui'
+
+const TRANSLATION_LOAD_TIMEOUT = 10_000
+let translationsPromise
 
 export default function translationPlugin(app) {
 	app.config.globalProperties.__ = translate
 	window.__ = translate
-	if (!window.translatedMessages) fetchTranslations()
```
raizasafeel, line 6: The plugin no longer loads translations itself, so the sub-apps in `utils/assignment.js` now depend on `main.js` having already initialized them. Is that dependency intentional?

frontend/src/translation.js
germanegomez: Fixed. Startup now reads the language-scoped IndexedDB cache and can mount with cached translations immediately. The request still refreshes the dictionary in the background, preserving the previous cache-and-refresh behavior.

frontend/src/translation.js
germanegomez: Fixed. Timeout continues with source messages without setting window.translatedMessages. The pending request can still finish; when it does, it updates the dictionary and reactive labels. A later load also reuses that pending request.

frontend/src/translation.js
germanegomez: Yes, for this entry point. main.js awaits loadTranslations() before mounting. The EditorJS tools and their Vue sub-apps are created from BlockEditor.vue after the root app mounts, so they share the loaded dictionary. The plugin only installs __ in each app context.

frontend/src/translation.js
```diff
@@ -28,13 +34,80 @@ function translate(message) {
...
+	let timeoutId
+	const timeout = new Promise((resolve) => {
+		timeoutId = window.setTimeout(() => {
+			console.warn(
+				'Translation startup timed out; using source messages.',
+			)
+			resolve({})
+		}, TRANSLATION_LOAD_TIMEOUT)
+	})
+	const available = Promise.any([cached, networkPromise]).catch((error) => {
```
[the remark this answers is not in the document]
germanegomez: Valid, fixed. The loader now selects the first successful source using standard Promise.then handlers, with the same bounded timeout and fallback if both sources fail. It no longer uses Promise.any.

germanegomez:
I rebuilt this branch after revisiting the root cause and the earlier review feedback.

The previous implementation synchronized bootstrap with translation loading, which put the resource on the critical path and required timeout and manual cache coordination. The updated commit keeps Vue mounting immediately, preserves the native `createResource` cache, and makes the shared translation catalog reactive instead. The main frontend and EditorJS sub-apps now observe the same singleton state.

Focused tests, the full 1,645-test frontend suite, formatting, and ESLint pass. The production build remains blocked by `Missing "./code-editor" specifier in "frappe-ui" package`; I reproduced the identical failure from clean `upstream/develop`, so it is not introduced by this PR. The new GitHub Actions runs are awaiting maintainer approval.

PR #2739, opened 2026-09-14
fix: Russian "Checkout" wording implied payment

ZubikIT:
Follow-up to #2733, addressing the review note left there.

`Checkout the course` / `Checkout the batch` were translated as **«Оформить»**, which in Russian reads as *pay for* / *place an order*. These strings label links in notification emails that simply open the published course or batch page — including for free content — so the wording promised a checkout that never happens.

**«Посмотреть»** (*view*) matches what the link actually does.

Two lines.


raizasafeel:
Thanks for the fix! Merged!

PR #2740, opened 2026-09-14
fix: translate the labels in the user menu

ZubikIT:
## The problem

Seven entries in the user dropdown are plain strings rather than `__('...')`, so they stay English no matter how complete a translation is:

```js
label: 'My Profile',
label: 'Toggle Theme',
label: 'Settings',
label: 'Clear Demo Data',
label: 'Login to Frappe Cloud',
label: 'Log out',
label: 'Log in',
```

"Toggle Theme" sitting in English next to a fully translated sidebar is the most visible case — the menu is the first thing a user opens.

## Why wrapping is safe here

`userDropdownOptions` is a `computed`, evaluated at render time, when `translationPlugin` has already installed `window.__`. The identical change inside a top-level constant would throw `__ is not a function` at module load — worth keeping in mind for the nav link constants elsewhere, which is why this PR does not touch them.

## Translations

Four labels resolve from frappe's own dictionary once wrapped. The other four had nowhere to resolve from: without `__()` the extractor never saw them, so no PO entries existed at all. Added, with Russian translations.

`msgfmt --check` passes.


raizasafeel:
Can you raise the PR again with just the russian translations? The fix for translations in the menu is already raised in this PR: #2737

PR #2741, opened 2026-09-14
feat: show which topic a question came from

ZubikIT:
## Why

A quiz that draws from a large bank mixes questions from every module, and the question header gives no hint which one is being asked about. The learner only finds out where they are weak from the final summary — too late to slow down on the topic they are shaky in.

This is most noticeable with `shuffle_questions` + `limit_questions_to`, where each attempt is a fresh sample across the whole bank.

## What

- an optional `topic` field on **LMS Question**
- rendered in the question header, opposite the marks

Empty by default, so existing quizzes look exactly as they do today.

The value goes through `__()`: a bank filled in by a course author is likely to carry topic names worth translating.

## Note

I am using this downstream through a Custom Field plus a small frontend patch, which works but means every author who wants it has to rediscover it. Upstreaming the field seemed more useful than keeping the workaround. Happy to adjust the placement or the label if you would rather have it elsewhere — for instance as a Link to a proper Topic doctype, if that fits your plans better than free text.


raizasafeel:
The use case makes sense but there needs to be proper ui for this. Could you first raise this as a feature request and then raise the PR against it with proper images?

PR #2743, opened 2026-09-14
feat: show chapter title in lesson form

### LINKED ISSUE #635: UX: show chapter title in lesson

Sandeepambala:
Fixes #635

## Why

A lesson is only reachable through its chapter, but the lesson form never showed which chapter it belonged to. With similar lesson names across chapters there was no way to tell which chapter's lesson was open — the outline sidebar has the answer, the editor itself does not.

## What

Show the chapter title above the lesson title.

No API change needed. `lms.lms.utils.get_lesson_creation_details` already returns `chapter.title`; the form was simply not reading it.

```js
const chapterTitle = computed(() => lessonDetails.data?.chapter?.title)
```

It renders only when a title is present, so the layout is unchanged if the chapter has no title.

## Tests

`frontend/src/tests/lessonChapterTitle.test.ts`, using the existing `lessonFormTeardown.test.ts` harness:

- shows the chapter the lesson belongs to
- shows it above the lesson title, not in place of it
- renders nothing when the chapter has no title

All three fail without the change.

Full suite run before and after — identical apart from the new tests. The 35 pre-existing file failures on `develop` (settings, profile, raven, multiLink) are unrelated and reproduce on a clean checkout.

|  | Test files | Tests |
|---|---|---|
| `develop` | 35 failed / 114 passed | 12 failed / 1199 passed |
| this branch | 35 failed / 115 passed | 12 failed / 1202 passed |

## Note

`LessonForm.vue` is not Prettier-clean on `develop`, so I did not run Prettier over the whole file — it would bury this change in unrelated reformatting. The added lines match Prettier's output, and the new test file is Prettier-clean.


raizasafeel:
Right now since you have the lesson editor on the right side indicating your current lesson, this shouldn't be a problem. So closing the PR, thanks for the contribution though!

PR #2746, opened 2026-09-15
chore: upgrade frappe-ui to 1.0.0-beta.69

raizasafeel:
## Summary

- Upgrades `frappe-ui` from `1.0.0-beta.29` to `1.0.0-beta.69` and moves the LMS frontend onto frappe-ui's v1 component API.
- Also bumps `lucide-static` from `1.17.0` to `^1.31.0`, which the new icon plugin needs.
- Fixes what the upgrade broke along the way: clipped tab icons, an overlapping Chapters header, a page-length dropdown with no options, and a lost table row height.
- Picks up the new espresso v2 colour palette: badges move from orange to amber, corner radii get numbered names.
- 218 files, 20 commits. 1694/1694 frontend tests pass, `yarn build` exits 0.

## Why

- We were 40 beta releases behind. frappe-ui v1 renamed props, slots and import paths across most of the component set, so the gap only gets more expensive to close.
- Staying behind also blocks the newer component families (charts, editor) that ship only in recent betas.

## Breaking changes

Anyone with custom frontend code or an in-flight frontend PR will hit these. What changed and what to do:

- **`frappe-ui/frappe` imports** — now come from `@framework/ui`.
- **Parked component families** — `Calendar`, the chart components, the `ListView` family and `TextEditor` now import from `frappe-ui/experimental` rather than the package root.
- **Popover-family slot API** — `toggle(flag)` is now `setOpen(value)`. Affects `Popover`, `Dropdown`, `DatePicker`, `Combobox`, `Select`, `MultiSelect`, `ContextMenu` and `HoverCard`.
- **`FileUploader`** — takes flat props instead of a nested `uploadArgs` object.
- **Tooltip and HoverCard delays are now milliseconds, not seconds.** This one is worth reading twice: `:hoverDelay="1.5"` still compiles, still runs, and now means 1.5 milliseconds instead of 1.5 seconds. Nothing warns you. Multiply every existing delay by 1000.
- **Design tokens** — removed radius aliases were renamed to numbered tokens, and the chromatic ink tokens shifted. Custom components using the old names will render with no radius rather than error.

## What's in it

- Dependency bump plus the build config it touches (vite, tailwind, `main.js`).
- `frappe-ui/frappe` imports moved to `@framework/ui`; parked families moved to `frappe-ui/experimental`.
- Overlays moved to the v1 API: Dialog, Tooltip, Popover, Dropdown, HoverCard, SettingsDialog.
- `FileUploader` on flat props, with an audit of every upload site to confirm each file still gets the public/private setting it had before.
- Token codemods: radius aliases renamed, ink tokens shifted, and the token parser taught to read OKLCH — it assumed hex, and the new palette ships OKLCH. Its alpha channel was also being computed gamma-encoded instead of linearly.
- Badges orange to amber, alerts stateless, tabs recomposed, shortcut combos on the v1 format.
- `debounce()` gained `.cancel()`, now used to drop the command palette's in-flight search when the query changes.
- Test mocks updated across 92 files; Playwright e2e selectors repaired.
- Tab strip fixes on course and batch detail pages (see below).
- Second bump, beta.65 to beta.69, with the `setOpen` rename and the delay unit conversion.

## Follow up
- More components are going to be ported from fui, this is simply basic migration of apis and colors


raizasafeel:
@greptile-apps

PR #2762, opened 2026-09-16
fix(profile): stop marking the profile image as mandatory

FawazAlhafiz:
Follow-up to #2656: the Profile Image label still shows the required asterisk, even though a profile saves fine without an image.

### What happens
1. Open your profile and click **Edit Profile**.
2. Look at the **Profile Image** label.
3. **It's marked required**, but saving without an image works.

profile-edit-image-still-mandatory

### Why
#2656 removed `:required="true"` from the `<Uploader>` in `ProfileEditForm.vue` so that only First Name is required. But `Uploader.vue` sets `required: true` as a default in `withDefaults`. Leaving the prop off doesn't make it `false`; the default applies instead, so the asterisk stayed.

Every other caller sets the prop explicitly (`BatchForm.vue` and `NewCourseForm.vue` pass `:required="false"`, `JobForm.vue` passes `:required="true"`). The profile form is the only one that relies on the default.

### Fix
Pass `:required="false"` explicitly.


FawazAlhafiz:
hello @raizasafeel could you take a look at this!
also I wonder if the **Uploader** it self should declare default required as False? since the usual rule is that a Boolean you leave off means OFF.

raizasafeel:
> hello @raizasafeel could you take a look at this! also I wonder if the **Uploader** it self should declare default required as False? since the usual rule is that a Boolean you leave off means OFF.

Yes, that's the proper fix, but for now, this seems fine. I will be redoing this soon so will change the defaults across the app then

